PR #3, opened 2022-03-17
Fix nvidia operator config path

m-bull:
- Force the nvidia gpu operator to create its own containerd config file at `/etc/containerd/conf.d`
- Ensure that `/etc/containerd/conf.d` is created on all clusters
- Enable the nvidia gpu operator exporter

### DIFF
--- a/charts/openstack-cluster/templates/_helpers.tpl
+++ b/charts/openstack-cluster/templates/_helpers.tpl
@@ -97,7 +97,6 @@ mirrors and additional packages.
 {{- with $kubeadmConfigSpec }}
 {{- toYaml . }}
 {{- end }}
-{{- if or $registryMirrors $files }}
 files:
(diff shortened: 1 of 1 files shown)


charts/openstack-cluster/templates/_helpers.tpl
```diff
@@ -99,6 +99,14 @@ mirrors and additional packages.
 {{- end }}
 {{- if or $registryMirrors $files }}
```
mkjpryor, line 100: @m-bull 

We want this file to always be created, but it is inside an `if` statement. You need to do some shuffling.

mkjpryor (changes requested):
Need to move an `if` statement to get the correct logic.

mkjpryor:
LGTM. Let's give it a go.

PR #5, opened 2022-03-21
Add ServiceMonitor for Nvidia DCGM exporter

m-bull:
The GPU operator doesn't deploy one yet (https://github.com/NVIDIA/gpu-operator/issues/305), so Prometheus doesn't know about the metrics endpoint exposed by the nvidia-dcgm-exporter. This adds config for a ServiceMonitor, gated behind `.Values.nvidiaGpuOperator.enabled`.

Also brings a small fix for Prometheus datasource names that breaks the dcgm-exporter-grafana-dashboard.

### DIFF
--- a/charts/cluster-addons/grafana-dashboards/nvidia-dcgm-exporter-dashboard_rev2.json
+++ b/charts/cluster-addons/grafana-dashboards/nvidia-dcgm-exporter-dashboard_rev2.json
@@ -1,7 +1,7 @@
 {
   "__inputs": [
     {
-      "name": "DS_PROMETHEUS",
+      "name": "Prometheus",
       "label": "Prometheus",
       "description": "",
       "type": "datasource",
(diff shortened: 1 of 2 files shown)


mkjpryor:
I think the alternative to this issue is to upgrade the gpu-operator chart from v1.9.1 to v1.10.0

mkjpryor:
@m-bull Perhaps you could try that and check it out?

xenon-dev:
Is there maybe official documentation somewhere, which gpu-operator umbrella chart works with which kube-prometheus-stack and which Grafana Version?

We tried to work with GPU-Operator v1.10.0 (why isn´t this semver??) and kube-prometheus-stack 34.1.1. and had no luck finding the dcgm exporter in the service monitor. Nothing in the offical docs state, that you have to add a custom service monitoring for dcgm exporter nor that currently all existing dashboards seem to be broken.

We will implement the changes proposed in this pr ourselves (no time to wait) and see if that will to the trick, otherwise there will only be the option to install something like HashiCorp Consul for a better service monitoring solution, but what else??

PR #11, opened 2022-10-25
[ccm] Use clusterName in generated cloud-config

yankcrime:
If the entry in the cloud-config file doesn't match the name of the cluster, then CCM fails to start and throws an error.  To fix this, we use clusterName if specified, otherwise fall back to the previous default of `openstack`.

### DIFF
--- a/charts/cluster-addons/templates/openstack/cloud-config.yaml
+++ b/charts/cluster-addons/templates/openstack/cloud-config.yaml
@@ -23,7 +23,7 @@ spec:
             [Global]
             use-clouds=true
             clouds-file=/etc/config/clouds.yaml
-            cloud=openstack
+            cloud={{ .Values.clusterName | default "openstack" }}
         {%- if "cacert" in cloud_identity.data %}
             ca-file=/etc/config/cacert
         {%- else %}


mkjpryor:
@yankcrime 

I don't think `.Values.clusterName` is the right thing to use here. It is intended to match the actual name of the Cluster API cluster and is passed to the OCCM/CSI Cinder to allow it to tag OS resources as belonging to the cluster. This needs to work even if every credential uses `openstack` in the `clouds.yaml`.

I think we need a new field under `.Values.openstack`, e.g. `.Values.openstack.cloudName`, that can be used to control this.

mkjpryor (changes requested):
See comment https://github.com/stackhpc/capi-helm-charts/pull/11#issuecomment-1300176216

yankcrime:
I think I conflated a couple of things here since we're using a [different Helm chart](https://github.com/eschercloudai/cluster-openstack) to deploy our clusters and I completely overlooked [where this value originally came from](https://github.com/eschercloudai/cluster-openstack/blob/692a374bb56f6e7253762e3957a695c817b2b73d/helm/cluster-openstack/values.yaml#L6).

I do think this should be configurable in this Helm chart though and not hardcoded as `openstack`, so I'll update this PR with your suggestion 👍

irt-shpc:
This PR is stale and the target branch is 4 years behind the stable branch. Closing PR

PR #12, opened 2022-10-25
[cilium] Switch the IPAM mode to the default

yankcrime:
The default IPAM mode with Cilium is [cluster scope](https://docs.cilium.io/en/stable/concepts/networking/ipam/), this commit switches back to this default so that there are no surprises when provisioning a cluster (i.e if you don't annotate nodes)


mkjpryor (approved):
LGTM

PR #22, opened 2022-12-05
Issue#21 Allow changing of project prefix via values.yaml

cunningr:
The current project label prefix of capi.stackhpc.com is embedded into the chart templates. We would like to move this prefix out into values.yaml.

Filing new issue for a PR to add:

Allow us the change/replace the capi.stackhpc.com label prefix via helm values


JohnGarbutt:
@scrungus can you check the permissions for testing PRs from other peoples forks please?

JohnGarbutt:
This looks like a reasonable tweak to me. I get that you want to brand those.

Some snags with the CI to work out when you are proposing a PR from a fork and so you don't have access (possibly correctly) to the clouds.yaml we are using for testing. Hopefully a fix is incoming on that.

cunningr:
"I get that you want to brand those." - It's more that we have an existing prod deployment that we are trying to integrate this with.  We have a number of other changes that I'll submit at some point that mainly add some additional template params.  If there is anything that doesn't make sense to be included, that's ok, just want to contribute back what we can.

scrungus:
I've made the changes to the CI on main, I think @cunningr you'd have to cherry pick latest commit / rebase on your fork so we can see it in this PR?

scrungus (approved):

PR #32, opened 2023-01-16
Allow the specification of additional storage classes

m-bull:
Renames the `storageClass` to `defaultStorageClass`, and adds an `additionalStorageClasses` variable.

Useful for clouds that have multiple volume types with particular performance characteristics.

### DIFF
--- a/charts/cluster-addons/templates/openstack/csi-cinder.yaml
+++ b/charts/cluster-addons/templates/openstack/csi-cinder.yaml
@@ -52,7 +52,7 @@ spec:
     - secret:
         name: {{ include "cluster-addons.componentName" (list . "csi-cinder") }}-config
         key: overrides
-{{- if .Values.openstack.csiCinder.storageClass.enabled }}
+{{- if .Values.openstack.csiCinder.defaultStorageClass.enabled }}
 ---
 apiVersion: addons.stackhpc.com/v1alpha1
 kind: Manifests
@@ -66,15 +66,13 @@ spec:
   releaseName: csi-cinder-storageclass
   manifestSources:
     - template: |
-        {{- with .Values.openstack.csiCinder.storageClass }}
+        {{- with .Values.openstack.csiCinder.defaultStorageClass }}
         apiVersion: storage.k8s.io/v1
         kind: StorageClass
         metadata:
           name: {{ .name }}
-          {{- if .isDefault }}
           annotations:
             storageclass.kubernetes.io/is-default-class: "true"
-          {{- end }}
         provisioner: cinder.csi.openstack.org
         parameters:
           availability: {{ .availabilityZone }}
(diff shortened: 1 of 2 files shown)


scrungus (approved):

JohnGarbutt (approved):

heytrav:
We would be keen this feature.

mkjpryor (changes requested):
My only comment is can we do all the storage classes as different entries in a single manifests resource please?

charts/cluster-addons/templates/openstack/csi-cinder.yaml
```diff
@@ -89,4 +87,38 @@ spec:
...
+          name: {{ $rangeItem.name }}
+        provisioner: cinder.csi.openstack.org
+        parameters:
+          availability: {{ $rangeItem.availabilityZone }}
+          {{- with $rangeItem.volumeType }}
+          type: {{ . }}
+          {{- end }}
+        reclaimPolicy: {{ $rangeItem.reclaimPolicy }}
+        allowVolumeExpansion: {{ $rangeItem.allowVolumeExpansion }}
+        volumeBindingMode: WaitForFirstConsumer
```
heytrav, line 118: Could we make this configurable? (same in `additionalStorageClasses`)

```yaml
volumeBindingMode: {{ .volumeBindingMode }}
```

charts/cluster-addons/templates/openstack/csi-cinder.yaml
```diff
@@ -89,4 +87,38 @@ spec:
...
+        apiVersion: storage.k8s.io/v1
+        kind: StorageClass
+        metadata:
+          name: {{ $rangeItem.name }}
+        provisioner: cinder.csi.openstack.org
+        parameters:
+          availability: {{ $rangeItem.availabilityZone }}
+          {{- with $rangeItem.volumeType }}
+          type: {{ . }}
+          {{- end }}
```
heytrav, line 115: Please add in `fstype` here (same in `additionalStorageclasses`)
```yaml
          {{- with .fstype }}
          fstype: {{ . }}
          {{- end }}

```

m-bull:
I'll try and fix this up this week - thanks for the reviews @heytrav and @mkjpryor

heytrav:
> My only comment is can we do all the storage classes as different entries in a single manifests resource please?

Just to clarify, would the idea be to put all the storageclasses into a single list and use an attribute like `default: true` to distinguish the default storageclass or would the helm chart still expect a `defaultStorageClass` and a list of `additionalStorageClasses`?

mkjpryor:
> > My only comment is can we do all the storage classes as different entries in a single manifests resource please?
> 
> Just to clarify, would the idea be to put all the storageclasses into a single list and use an attribute like `default: true` to distinguish the default storageclass or would the helm chart still expect a `defaultStorageClass` and a list of `additionalStorageClasses`?

I actually only meant to use a single `Manifests` resource for all the storage classes, not change the values interface, for now.

If we were to reduce it to just a single list I think we would want to refactor to have a `storageClassDefaults` + `storageClasses` list, and merge them - similar to how the node groups work - in order to retain control over the defaults in a way that overrides nicely. What I mean by that is overriding individual properties of specific list items is not really possible in Helm, so it would prevent us specifying sensible defaults unless we do this refactor.

So I would prefer that, if we do that, it comes in a separate patch.

heytrav:
> > > My only comment is can we do all the storage classes as different entries in a single manifests resource please?
> > 
> > 
> > Just to clarify, would the idea be to put all the storageclasses into a single list and use an attribute like `default: true` to distinguish the default storageclass or would the helm chart still expect a `defaultStorageClass` and a list of `additionalStorageClasses`?
> 
> I actually only meant to use a single `Manifests` resource for all the storage classes, not change the values interface, for now.

Thanks. That's what i expected but just wanted to confirm. I'm working on a function in magnum-capi-helm to pull the values out of cinder.

mkjpryor:
@heytrav That is good to hear because I wanted to implement that!

PR #42, opened 2023-02-28
Add additional dashboards for Calico + systemd logs

mkjpryor:


charts/cluster-addons/values.yaml
```diff
@@ -168,54 +168,29 @@ ingress:
...
   # Indicates if the cluster monitoring should be enabled
   enabled: false
-  # labels to be added to ServiceMonitor resources
-  # must match labels from .serviceMonitorSelector.matchLabels
-  # field of Prometheus resource created by kube-prometheus-stack
-  # in order for Prometheus to scrape metrics from the services
-  serviceMonitorLabels:
-    release: kube-prometheus-stack
-  # Size of the volume in GB to provision on the target cloud for persistent storage of prometheus data
-  prometheusVolumeCapacity: 10
```
sd109, line 178: I'll need to update this in my azimuth-ops PR, but I think removing it makes sense here so that's fine.

charts/cluster-addons/values.yaml
```diff
@@ -224,9 +199,12 @@ monitoring:
       version: 2.8.9
     release:
       namespace: monitoring-system
-      values: {}
-  # Size of the volume in GB to provision on the target cloud for persistent storage of loki data
-  lokiVolumeCapacity: 10
```
sd109, line 229: As above

sd109 (approved):

charts/cluster-addons/values.yaml
mkjpryor: I have a new azimuth-ops PR that I started from your branch, so don't worry about updating

PR #47, opened 2023-04-20
Add nginx cert alerting rules

sd109:


mkjpryor (changes requested):
This needs to be a `Manifests` resource that installs the rule onto the remote cluster. This will install the rule on the CAPI management cluster, which is not what we intend.

The `Manifests` resource should live here: https://github.com/stackhpc/capi-helm-charts/blob/main/charts/cluster-addons/templates/ingress-nginx.yaml#L78

You will probably also need to add `ruleSelectorNilUsesHelmValues: false` to the monitoring deployment, similar to here: https://github.com/stackhpc/capi-helm-charts/blob/main/charts/cluster-addons/templates/monitoring/kube-prometheus-stack.yaml#L72

PR #51, opened 2023-06-13
Add support for uploading required images for CI

mkjpryor:
Also adds a test for upgrading from the latest tag.


.github/actions/ensure-image/action.yml
```diff
@@ -0,0 +1,38 @@
...
+  Ensures that the specified image exists in the target OpenStack cloud.
+
+inputs:
+  os-client-config-file:
+    description: The path of the OpenStack clouds file
+    required: true
+    default: ./clouds.yml
+  os-cloud:
+    description: The name of the cloud within the OpenStack clouds file
+    required: true
```
scrungus, line 13: same for this

.github/actions/ensure-image/action.yml
```diff
@@ -0,0 +1,38 @@
...
+    default: openstack
+  images-base-url:
+    description: The base URL to use for downloading images
+    required: true
```
scrungus, line 17: same for this

.github/actions/ensure-image/action.yml
```diff
@@ -0,0 +1,38 @@
+name: Ensure OpenStack image
+
+description: >-
+  Ensures that the specified image exists in the target OpenStack cloud.
+
+inputs:
+  os-client-config-file:
+    description: The path of the OpenStack clouds file
+    required: true
```
scrungus, line 9: should this be required if there is a default?

.github/actions/upgrade-and-test/action.yml
```diff
@@ -0,0 +1,157 @@
...
+  Run a Helm upgrade using the specified values, wait for the cluster to
+  become ready and run Sonobuoy against it
+
+inputs:
+  name:
+    description: The name of the cluster
+    required: true
+  os-client-config-file:
+    description: The path of the OpenStack clouds file
+    required: true
```
scrungus, line 13: should this be required or default - same goes for the rest of the inputs in this file that have both

.github/workflows/main.yaml
scrungus, line 1: this seems redundant if all pushes to main come through a pr, which carries out all these checks anyway?

.github/workflows/test.yaml
```diff
@@ -1,39 +1,40 @@
...
   workflow_call:
+    inputs:
+      images:
+        type: string
+        description: JSON-encoded dictionary of images and versions
+        required: true
+      tests-full:
+        type: boolean
+        description: Indicates whether to run the full test suite or just a smoke test
+        required: true
```
scrungus, line 12: same here

scrungus (commented):
couple things but otherwise LGTM

.github/actions/upgrade-and-test/action.yml
mkjpryor: I think required + default prevents the input from being set to `null` in a caller.

.github/actions/ensure-image/action.yml
mkjpryor: See below

.github/actions/ensure-image/action.yml
mkjpryor: See below

.github/actions/ensure-image/action.yml
mkjpryor: See below

.github/workflows/main.yaml
mkjpryor: We did decide that in the past, but I changed my mind.

I think I like running an image sync + smoke test on main just in case there is an emergency bugfix or something that needs to be force pushed is sensible.

.github/workflows/test.yaml
mkjpryor: See above

.github/actions/upgrade-and-test/action.yml
scrungus: ah, haven't seen that before. Good to know!

.github/workflows/main.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+  ensure_capi_images:
+    needs: [lint]
+    uses: ./.github/workflows/ensure-capi-images.yaml
+    secrets: inherit
+
+  test:
+    needs: [sync_images, ensure_capi_images]
+    uses: ./.github/workflows/test.yaml
+    secrets: inherit
```
scrungus, line 28:
```suggestion
    secrets: inherit
    if: ${{ github.event.forced }}
```

.github/workflows/main.yaml
scrungus: In that case, I think something like my below code suggestions is possible. Also see https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#push

mkjpryor: I get it. Something in me just likes the idea of smoke testing every commit to main just to be sure.

scrungus: Yeah, can do that then

scrungus:
LGTM

scrungus (approved):

PR #53, opened 2023-07-05
Flatcar integration

heytrav:
* Modify openstack-cluster templates to integrate ignition components for Flatcar
* Smaller changes related to addon defaults

### DIFF
--- a/charts/openstack-cluster/templates/_helpers.tpl
+++ b/charts/openstack-cluster/templates/_helpers.tpl
@@ -203,6 +203,7 @@ files:
       # This file is created by the capi-helm-chart to ensure that its parent directory exists
     owner: root:root
     permissions: "0644"
+{{- if ne .Values.osDistro "flatcar" }}
   - path: /etc/containerd/config.toml
     content: |
       [plugins."io.containerd.grpc.v1.cri".registry]
@@ -210,6 +211,7 @@ files:
     owner: root:root
     permissions: "0644"
     append: true
+{{- end }}
 {{- with .Values.registryMirrors }}
 {{- range $registry, $registrySpec := . }}
   - path: /etc/containerd/certs.d/{{ $registry }}/hosts.toml
@@ -274,4 +276,46 @@ Produces the spec for a KubeadmConfig object.
     $kubeadmConfigSpec |
   include "openstack-cluster.mergeConcatMany"
 }}
+{{- end }}
+
+{{/*
+Produces the spec for an Ignition based OS specific KubeadmConfig object conditional on osDistro set to "flatcar".
+*/}}
+{{- define "openstack-cluster.flatcarKubeadmConfigSpec" -}}
+initConfiguration:
+  nodeRegistration:
+    name: ${COREOS_OPENSTACK_HOSTNAME}
+joinConfiguration:
+  nodeRegistration:
+    name: ${COREOS_OPENSTACK_HOSTNAME}
+preKubeadmCommands:
+  - export COREOS_OPENSTACK_HOSTNAME=${COREOS_OPENSTACK_HOSTNAME%.*}
+  - envsubst < /etc/kubeadm.yml > /etc/kubeadm.yml.tmp
+  - mv /etc/kubeadm.yml.tmp /etc/kubeadm.yml
+format: ignition
+ignition:
+  containerLinuxConfig:
+    additionalConfig: |
+      systemd:
+        units:
+        - name: coreos-metadata-sshkeys@.service
+          enabled: true
+        - name: kubeadm.service
+          enabled: true
+          dropins:
+          - name: 10-flatcar.conf
+            contents: |
+              [Unit]
+              Requires=containerd.service coreos-metadata.service
+              After=containerd.service coreos-metadata.service
+              [Service]
+              EnvironmentFile=/run/metadata/flatcar
+{{- end }}
+
+{{- define "openstack-cluster.osDistroKubeadmConfigSpec" }}
+{{- $ctx := index . 0 }}
+{{- $osDistro := $ctx.Values.osDistro }}
+{{- if eq $osDistro "flatcar" }}
+{{- include "openstack-cluster.flatcarKubeadmConfigSpec" $ctx }}
+{{- end }}
 {{- end }}
--- a/charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
+++ b/charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
@@ -6,9 +6,13 @@ joinConfiguration: {{ include "openstack-cluster.nodeRegistration.nodeLabels" . 
 {{- end }}
 
 {{- define "openstack-cluster.controlplane.kubeadmConfigSpec.kubeProxyConfiguration" -}}
-{{- with .kubeProxyConfiguration }}
+{{- $ctx := index . 0 }}
+{{- $osDistro := index . 1 }}
+{{- $kubeProxyConfigurationPath := ternary "/etc/kube-proxy-configuration.yaml" "/run/kubeadm/kube-proxy-configuration.yaml" (eq $osDistro "flatcar") }}
+{{- $kubeadmPath := ternary "/etc/kubeadm.yml" "/run/kubeadm/kubeadm.yaml" (eq $osDistro "flatcar") }}
+{{- with $ctx.kubeProxyConfiguration }}
 files:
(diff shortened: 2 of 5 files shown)


mkjpryor:
I don't like how Flatcar feels like a bolt-on here. If we are going to support it, I feel we should refactor the charts so that there is an OS abstraction that Ubuntu and Flatcar both use.

mkjpryor:
I think there is a nice way to do it by utilising template definitions and passing around the name of a template to use, but I haven't got it fully figured out yet.

heytrav:
> I think there is a nice way to do it by utilising template definitions and passing around the name of a template to use, but I haven't got it fully figured out yet.

I wonder if something similar could be done for #91. Seems like some sort of abstraction for adding patches to the kube-apiserver config could also be useful.

mkjpryor:
@heytrav I was thinking the same. Let me have a bit of a think about how it would work.

heytrav:
@mkjpryor based the changes you made in #130 it looks like I can just override `controlPlane.kubeadmConfigSpec` and `nodeGroupDefaults.kubeadmConfigSpec` to add the necessary ignition components in a user defined `-f values.yaml`.  I'm still experimenting with it to be sure.

For the purpose of this PR I'm wondering if you want to 

1. Provide any sort of conditional blocks in `charts/openstack-cluster/values.yaml` to switch between cloud-init or ignition based systems OR
2. Leave that out completely and leave it up to the user to override with the ignition components?

For the latter case I would probably just close this PR. We'll just use our own with the ignition setup regardless.

heytrav:
I've refactored the original PR to remove most of the changes I had made. In the current state it is possible to build Flatcar clusters with this code and providing customised `values.yaml`.

However as mentioned in #138  it was necessary to change the last command added to `preKubeadmCommands` because it references a folder that does not exist on Flatcar and causes the build to fail. Rather than remove it completely I added a separate definition that uses a different path when `ignitionBasedOS: true`.

To resolve #139 I changed the destination of the additional containerd config to create a new file under `/etc/containerd/config.d/`. Not sure if this is best approach but for now it at least doesn't clobber the containerd configuration

charts/openstack-cluster/templates/_helpers.tpl
```diff
@@ -203,10 +203,11 @@ files:
...
     owner: root:root
     permissions: "0644"
-  - path: /etc/containerd/config.toml
+  - path: /etc/containerd/config.d/containerd-certs.toml
     content: |
-      [plugins."io.containerd.grpc.v1.cri".registry]
-      config_path = "/etc/containerd/certs.d"
+      [plugins]
+        [plugins."io.containerd.grpc.v1.cri".registry]
+        config_path = "/etc/containerd/certs.d"
```
mkjpryor, line 210: We actually tried this, but it doesn't work because of this:

https://github.com/containerd/containerd/issues/5837

That is why we used `append: true` here, which clearly doesn't work in Ignition.

So this change will need to be reverted and we will need to think of another solution. It would be nice to be able to do this in Flatcar also (i.e. support registry mirrors).

charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
```diff
@@ -21,6 +21,22 @@ preKubeadmCommands:
...
+      ---
+      apiVersion: kubeproxy.config.k8s.io/v1alpha1
+      kind: KubeProxyConfiguration
+      {{- toYaml . | nindent 6 }}
+    owner: root:root
+    permissions: "0644"
+preKubeadmCommands:
+  - cat /etc/kube-proxy-configuration.yaml >> /run/kubeadm.yml
+{{- end }}
+{{- end }}
```
mkjpryor, line 38: Can this be done by just gating the path and command on `.Values.ignitionBasedOS`? I'm not that keen on the duplication.

charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
```diff
@@ -47,6 +63,22 @@ spec:
...
+          (include "openstack-cluster.kubeadmConfigSpec" (list . .Values.controlPlane.kubeadmConfigSpec) | fromYaml) 
+          (include "openstack-cluster.controlplane.kubeadmConfigSpec.ignitionKubeProxyConfiguration" .Values.controlPlane.kubeadmConfigSpec | fromYaml) |
+        include "openstack-cluster.mergeConcatMany" |
+        fromYaml
+      )
+      "kubeProxyConfiguration" |
+    toYaml |
+    nindent 4
+  }}
+  {{- else }}
```
mkjpryor, line 81: Which would mean this is not needed. We would need to modify the call below to pass in the whole context.

mkjpryor (changes requested):

heytrav:
> @heytrav
> 
> I think in the short- to medium-term it would be nice to have the Flatcar support be native to the charts, so that we only require the user to flip a value in the charts between `flatcar` and `ubuntu`. However if you think this is the minimum set of changes that enable the use of Flatcar and you don't have the time or inclination to look at the required changes for that, I am happy to merge (subject to a full review and passing the test suite).
> 
> Let me know what you prefer.

I actually would like to have ignition based OS (i.e. flatcar) be completely handled by the chart as you suggest. However I wasn't sure how you preferred this to work. 

I will need to experiment a bit to add this into the chart. If you have any suggestions, I'm happy to follow your lead.

charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
heytrav: Inside the block I was unable to access the  `.Values.ignitionBasedOS` flag. The helm chart would fail to render because the `.ignitionBasedOS` variable didn't exist. Maybe that happens because it is in the `.kubeProxyConfiguration` context?  I'm not sure. 

Perhaps there's a trick to getting the root level objects

heytrav:
Most recent commit works for me locally (I can build a Flatcar cluster).

@mkjpryor please let me know if this is more in line with what you were thinking.

charts/openstack-cluster/templates/_helpers.tpl
```diff
@@ -203,13 +203,15 @@ files:
       # This file is created by the capi-helm-chart to ensure that its parent directory exists
     owner: root:root
     permissions: "0644"
+{{- if not (.Values.ignitionBasedOS | default false) }}
```
mkjpryor, line 206: I would prefer we just set `ignitionBasedOS: false` in the chart `values.yaml`, then we don't need the `default` here.

mkjpryor: Similarly elsewhere

charts/openstack-cluster/values.yaml
```diff
@@ -119,6 +119,39 @@ apiServer:
...
+          - name: kubeadm.service
+            enabled: true
+            dropins:
+            - name: 10-flatcar.conf
+              contents: |
+                [Unit]
+                Requires=containerd.service coreos-metadata.service
+                After=containerd.service coreos-metadata.service
+                [Service]
+                EnvironmentFile=/run/metadata/flatcar
```
mkjpryor, line 154: I was actually thinking more like all this stuff being in the `openstack-cluster.ignitionKubeadmConfigSpec`, so that we don't even have a separate variable here. This would be more consistent with the existing setup. Maybe that template should be called `openstack-cluster.flatcarKubeadmConfigSpec` in case there are other Ignition-based OSs that have different requirements to Flatcar, and instead of `.Values.ignitionBasedOS` we can have `.Values.osName` that defaults to `ubuntu`.

It might also involve moving anything that is Ubuntu-specific from `values.yaml` to a corresponding `openstack-cluster.ubuntuKubeadmConfigSpec` template.

The user can still customise the kubeadmConfigSpec for either Flatcar or Ubuntu via the existing hooks.

mkjpryor (changes requested):

charts/openstack-cluster/values.yaml
heytrav: Should we call it `osDistro` to align with the glance image convention?

heytrav:
> @heytrav
> 
> I added my thoughts in the form of review comments. Since we are adding support for a second OS, I would like to make sure that we can actually support N OSs going forward, just in case, so the suggestions I made reflect that.

pushed up some changes

mkjpryor (approved):
LGTM

PR #54, opened 2023-07-07
Make rootVolume dependent on disk size being specified

mkjpryor:
When used in Azimuth, this will allow the volume type and AZ to be specified globally or by the template while allowing the individual cluster to specify volume sizes without worrying about conditional merges.


mkjpryor:
@scrungus Prod :-)

scrungus (approved):
LGTM

PR #55, opened 2023-07-17
Use a dependencies file

mkjpryor:
Pull dependency versions from a new dependencies file that makes them (a) explicit instead of buried in CI files and (b) easier to update, possibly with automation, in the future.

We also add support for the new azimuth-images manifests. This allows us to just declare a version of the azimuth-images to use and pull the name/url/kubernetes_version out automatically.


scrungus (approved):
LGTM

PR #72, opened 2023-08-15
Use an app token so that automation PRs trigger workflows

mkjpryor:
This PR adds the ability to use a GitHub App token to authenticate PRs that are created to update dependencies. This means that those PRs will be able to trigger automation like the PR tests.


scrungus:
LGMT

PR #88, opened 2023-09-11
Sync mofed 5.7-1.0.2.0-ubuntu20.04-amd64

m-bull:
For clouds with older versions of the operator deployed


mkjpryor:
Image syncing has been abandoned for now

PR #90, opened 2023-09-14
Separate machine image for control plane and worker nodes

heytrav:
To satisfy the [version skew policy](https://kubernetes.io/releases/version-skew-policy/) it would be useful if 
control plane and machine deployment nodes could be updated separately. This would allow us to update the control plane first and then update the worker nodes.

This checks if machineImage and kubernetesVersion are defined for the separate control plane/nodegroup and falls back to the root one if not.

~~Allow separate (or no) imagePrefix for addons~~

### DIFF
--- a/charts/openstack-cluster/templates/control-plane/openstack-machine-template.yaml
+++ b/charts/openstack-cluster/templates/control-plane/openstack-machine-template.yaml
@@ -16,12 +16,16 @@ template:
     {{- if .Values.controlPlane.machineRootVolume.diskSize }}
     rootVolume: {{ toYaml .Values.controlPlane.machineRootVolume | nindent 6 }}
     {{- end }}
-    {{- if .Values.machineImageId }}
+    {{- if .Values.controlPlane.machineImageId }}
+    imageUUID: {{ .Values.controlPlane.machineImageId }}
+    {{- else if .Values.controlPlane.machineImage }}
+    image: {{ tpl .Values.controlPlane.machineImage . }}
+    {{- else if .Values.machineImageId }}
     imageUUID: {{ .Values.machineImageId }}
     {{- else if .Values.machineImage }}
     image: {{ tpl .Values.machineImage . }}
     {{- else }}
-    {{- fail "Either machineImage or machineImageId is required" }}
+    {{- fail "Either controlPlane.machineImageId, controlPlane.machineImage, machineImage or machineImageId is required" }}
     {{- end }}
     {{- with .Values.controlPlane.machineNetworking.ports }}
     ports: {{ toYaml . | nindent 6 }}
--- a/charts/openstack-cluster/templates/node-group/machine-deployment.yaml
+++ b/charts/openstack-cluster/templates/node-group/machine-deployment.yaml
@@ -47,7 +47,7 @@ spec:
         {{ $.Values.projectPrefix }}/node-group: {{ $nodeGroup.name }}
     spec:
       clusterName: {{ include "openstack-cluster.clusterName" $ }}
-      version: v{{ trimPrefix "v" $.Values.kubernetesVersion }}
+      version: v{{ $nodeGroup.kubernetesVersion | default $.Values.kubernetesVersion | trimPrefix "v" }}
       {{- with $nodeGroup.failureDomain }}
       failureDomain: {{ . }}
       {{- end }}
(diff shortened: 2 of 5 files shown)


charts/openstack-cluster/templates/node-group/openstack-machine-template.yaml
```diff
@@ -20,10 +20,12 @@ template:
     {{- end }}
     {{- if $ctx.Values.machineImageId }}
     imageUUID: {{ $ctx.Values.machineImageId }}
+    {{- else if $ctx.Values.nodeGroupDefaults.machineImage }}
```
mkjpryor, line 23: See above

charts/openstack-cluster/values.yaml
```diff
@@ -131,6 +131,10 @@ controlPlane:
   # For high-availability, this should be greater than 1
   # For etcd quorum, it should be odd - usually 3, or 5 for very large clusters
   machineCount: 3
+  # The kubernetes version for the control plane
+  kubernetesVersion:
+  # The image to use for control plane
+  machineImage:
```
mkjpryor, line 137: See above

charts/openstack-cluster/values.yaml
```diff
@@ -228,6 +232,9 @@ nodeGroupDefaults:
   failureDomain:
   # The flavor to use for machines in the node group
   machineFlavor:
+  # Default image id for nodeGroup hosts
+  machineImage:
+  kubernetesVersion:
```
mkjpryor, line 237: See above

mkjpryor (changes requested):
Just some minor changes. Basically, if we are going to do this for image name we should also do it for image ID, I think.

charts/openstack-cluster/templates/control-plane/openstack-machine-template.yaml
```diff
@@ -18,10 +18,12 @@ template:
     {{- end }}
     {{- if .Values.machineImageId }}
     imageUUID: {{ .Values.machineImageId }}
+    {{- else if .Values.controlPlane.machineImage }}
```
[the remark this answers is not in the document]
heytrav: I've added machineImageId to both control plane and nodeGroup sections

charts/openstack-cluster/templates/node-group/openstack-machine-template.yaml
```diff
@@ -18,12 +18,16 @@ template:
     {{- if $nodeGroup.machineRootVolume.diskSize }}
     rootVolume: {{ toYaml $nodeGroup.machineRootVolume | nindent 6 }}
     {{- end }}
-    {{- if $ctx.Values.machineImageId }}
+    {{- if $ctx.Values.nodeGroupDefaults.machineImageId }}
+    imageUUID: {{ $ctx.Values.nodeGroupDefaults.machineImageId }}
```
mkjpryor, line 22: This should be `$nodeGroup.machineImageId` rather than `$ctx.Values.nodeGroupDefaults.machineImageId` - see above.

charts/openstack-cluster/templates/node-group/machine-deployment.yaml
```diff
@@ -47,7 +47,7 @@ spec:
         {{ $.Values.projectPrefix }}/node-group: {{ $nodeGroup.name }}
     spec:
       clusterName: {{ include "openstack-cluster.clusterName" $ }}
-      version: v{{ trimPrefix "v" $.Values.kubernetesVersion }}
+      version: v{{ $.Values.nodeGroupDefaults.kubernetesVersion | default $.Values.kubernetesVersion | trimPrefix "v" }}
```
mkjpryor, line 50: This should be `$nodeGroup.kubernetesVersion` instead of `$.Values.nodeGroupDefaults.kubernetesVersion`.

This will pick up the value from `nodeGroupDefaults` if not overridden in the node group, but also allows having a separate image per node group if desired.

charts/openstack-cluster/templates/node-group/openstack-machine-template.yaml
```diff
@@ -18,12 +18,16 @@ template:
...
+    {{- else if $ctx.Values.nodeGroupDefaults.machineImage }}
+    image: {{ tpl $ctx.Values.nodeGroupDefaults.machineImage $ctx }}
```
mkjpryor, line 24: As above

mkjpryor (changes requested):
Just a couple more small changes. Sorry :-)

charts/openstack-cluster/templates/node-group/openstack-machine-template.yaml
heytrav: I assume that should also apply to `$ctx.Values.nodeGroupDefaults.machineImage` right below that?

heytrav: I changed the `$nodeGroup.machineImageId` and also `$nodeGroup.machineImage` as well.

mkjpryor (approved):
LGTM

Unfortunately, we don't have the ability to run the tests against forked repos ATM, so I will need to run the tests manually.

PR #93, opened 2023-09-19
Remove syncing of images

mkjpryor:
The image syncing turns out to be a massive PITA to maintain, and makes the automated updating of addon versions almost impossible to do.

So this patch removes the Skopeo image syncing in favour of configuring a mirror for Docker Hub in the CI.


scrungus (approved):
LGTM

PR #110, opened 2023-09-19
Update to CAPO 0.8.0 + v1alpha7

mkjpryor:


.github/actions/setup/action.yml
```diff
@@ -0,0 +1,115 @@
...
+    - name: Install clusterctl
+      shell: bash
+      run: >
+        curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/${CAPI_VERSION}/clusterctl-linux-amd64 -o clusterctl &&
+        sudo install -o root -g root -m 0755 clusterctl /usr/local/bin/clusterctl &&
+        clusterctl version
+      env:
+        CAPI_VERSION: ${{ steps.deps.outputs.cluster-api }}
+
+    - name: Check if Cluster API is already installed
```
scrungus, line 74: why would cluster-api already be installed?

.github/workflows/test.yaml
```diff
@@ -171,103 +54,218 @@ jobs:
...
-      #####
+      - name: Apply network configuration
+        run: echo "$NETWORKING" >> ./values.yaml
+        env:
+          NETWORKING: |
+            clusterNetworking:
+              internalNetwork:
+                networkFilter:
+                  tags: capi-helm-chart-ci
+        if: ${{ !inputs.tests-full }}
```
scrungus, line 65: why do we not want to create a new network if we're doing a quick test?

scrungus (approved):
Pretty neat!

.github/workflows/test.yaml
mkjpryor: Because the most common failure is quota error on routers :)

.github/actions/setup/action.yml
mkjpryor: I modified the chart upgrade test so that it also upgrades the dependencies, so the flow is:

1. Check out the last tag
1. Install the dependencies from there
2. Use the chart from there to make a cluster
3. Check out the code under test
4. Update the dependencies from there
5. Upgrade the cluster to the new chart

mkjpryor: That allows us to test moving a cluster from one CAPI or CAPO version to the next

PR #91, opened 2023-09-14
Add configuration for keystone webhook

heytrav:
* Add template for basic kube-apiserver webhook based authN/authZ 
* Add support for deploying sub chart based on https://github.com/heytrav/helm-charts/tree/main/charts/k8s-keystone-auth

PR #130, opened 2023-10-02
Add support for specifying node labels

mkjpryor:
This PR also adds a layering mechanism for building up the `KubeadmConfigSpec` that should help with #53 and #91.


charts/openstack-cluster/templates/_helpers.tpl
```diff
@@ -178,40 +219,59 @@ files:
...
-  - update-ca-certificates
-  {{- end }}
-  {{- if $additionalPackages }}
   - apt update -y
-  - apt install -y {{ join " " $additionalPackages }}
-  {{- end }}
-  {{- range $preKubeadmCommands }}
-  - {{ . }}
-  {{- end }}
+  - apt install -y {{ join " " . }}
```
m-bull, line 259: Can/should `apt install -y` be parameterised to support other package management commands (e.g. `dnf install -y`)?

m-bull (approved):

charts/openstack-cluster/templates/_helpers.tpl
mkjpryor: I was thinking that can follow in subsequent patches, with `openstack-cluster.kubeadmConfigSpec.additionalPackages.{ubuntu,rocky,...}`.

The layering implemented in this patch makes that easier in the future.

PR #168, opened 2023-12-04
Allow multiple storage classes

heytrav:
Rebase and small refactoring of #32

### DIFF
--- a/charts/cluster-addons/templates/openstack/csi-cinder.yaml
+++ b/charts/cluster-addons/templates/openstack/csi-cinder.yaml
@@ -55,7 +55,7 @@ spec:
     - secret:
         name: {{ include "cluster-addons.componentName" (list . "csi-cinder") }}-config
         key: overrides
-{{- if .Values.openstack.csiCinder.storageClass.enabled }}
+{{- if .Values.openstack.csiCinder.defaultStorageClass.enabled }}
 ---
 apiVersion: addons.stackhpc.com/v1alpha1
 kind: Manifests
@@ -72,15 +72,13 @@ spec:
   releaseName: csi-cinder-storageclass
   manifestSources:
     - template: |
-        {{- with .Values.openstack.csiCinder.storageClass }}
+        {{- with .Values.openstack.csiCinder.defaultStorageClass }}
         apiVersion: storage.k8s.io/v1
         kind: StorageClass
         metadata:
           name: {{ .name }}
-          {{- if .isDefault }}
           annotations:
             storageclass.kubernetes.io/is-default-class: "true"
-          {{- end }}
         provisioner: cinder.csi.openstack.org
         parameters:
           availability: {{ .availabilityZone }}
@@ -94,5 +92,39 @@ spec:
         allowedTopologies: {{ toYaml . | nindent 6 }}
         {{- end }}
         {{- end }}
+{{- end }}
+{{- range .Values.openstack.csiCinder.additionalStorageClasses }}
+{{- $rangeItem := . -}}
+{{- with $ }}
+---
+apiVersion: addons.stackhpc.com/v1alpha1
+kind: Manifests
+metadata:
+  name: {{ include "cluster-addons.componentName" (list . "csi-cinder-storageclass") }}
+  labels: {{ include "cluster-addons.componentLabels" (list . "csi-cinder-storageclass") | nindent 4 }}
+spec:
+  clusterName: {{ include "cluster-addons.clusterName" . }}
+  bootstrap: true
+  targetNamespace: {{ .Values.openstack.targetNamespace }}
+  releaseName: {{ $rangeItem.name }}-storageclass
+  manifestSources:
+    - template: |
+        apiVersion: storage.k8s.io/v1
+        kind: StorageClass
+        metadata:
+          name: {{ $rangeItem.name }}
+        provisioner: cinder.csi.openstack.org
+        parameters:
+          availability: {{ $rangeItem.availabilityZone }}
+          {{- with $rangeItem.volumeType }}
+          type: {{ . }}
+          {{- end }}
+        reclaimPolicy: {{ $rangeItem.reclaimPolicy }}
+        allowVolumeExpansion: {{ $rangeItem.allowVolumeExpansion }}
+        volumeBindingMode: WaitForFirstConsumer
+        {{- with $rangeItem.allowedTopologies }}
+        allowedTopologies: {{ toYaml . | nindent 6 }}
+        {{- end }}
+{{- end }}
 {{- end }}
 {{- end }}
--- a/charts/cluster-addons/templates/openstack/csi-cinder.yaml
+++ b/charts/cluster-addons/templates/openstack/csi-cinder.yaml
@@ -85,29 +85,17 @@ spec:
           {{- with .volumeType }}
           type: {{ . }}
           {{- end }}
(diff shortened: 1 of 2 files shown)


charts/cluster-addons/templates/openstack/csi-cinder.yaml
```diff
@@ -72,27 +72,55 @@ spec:
...
           {{- with .volumeType }}
           type: {{ . }}
           {{- end }}
+          {{- with .fstype }}
+          fstype: {{ . }}
+          {{- end }}
         reclaimPolicy: {{ .reclaimPolicy }}
         allowVolumeExpansion: {{ .allowVolumeExpansion }}
-        volumeBindingMode: WaitForFirstConsumer
+        volumeBindingMode: {{ .volumeBindingModea }}
```
mkjpryor, line 93: Typo

charts/cluster-addons/templates/openstack/csi-cinder.yaml
```diff
@@ -72,27 +72,55 @@ spec:
...
         {{- with .allowedTopologies }}
         allowedTopologies: {{ toYaml . | nindent 6 }}
         {{- end }}
         {{- end }}
+{{- $storeageClassIndex := 0 }}
+{{- range .Values.openstack.csiCinder.additionalStorageClasses }}
+{{- $rangeItem := . -}}
```
mkjpryor, line 100: I think you can do this as

```
{{- range $rangeItem := .Values.openstack.csiCinder.additionalStorageClasses }}
```

??

charts/cluster-addons/templates/openstack/csi-cinder.yaml
```diff
@@ -72,27 +72,55 @@ spec:
...
+          {{- end }}
         reclaimPolicy: {{ .reclaimPolicy }}
         allowVolumeExpansion: {{ .allowVolumeExpansion }}
-        volumeBindingMode: WaitForFirstConsumer
+        volumeBindingMode: {{ .volumeBindingModea }}
         {{- with .allowedTopologies }}
         allowedTopologies: {{ toYaml . | nindent 6 }}
         {{- end }}
         {{- end }}
+{{- $storeageClassIndex := 0 }}
```
mkjpryor, line 98: Typo

Also, what is this actually used for? As far as I can see it is incremented but never used.

charts/cluster-addons/templates/openstack/csi-cinder.yaml
```diff
@@ -72,27 +72,55 @@ spec:
...
+{{- range .Values.openstack.csiCinder.additionalStorageClasses }}
+{{- $rangeItem := . -}}
+{{- with $ }}
```
mkjpryor, line 101: Is this actually necessary? I don't see any use of `.Values`, `.Chart`, etc. inside the loop?

charts/cluster-addons/values.yaml
```diff
@@ -89,25 +88,32 @@ openstack:
...
+      # reclaimPolicy: Delete
+      # # Indicates if volume expansion is allowed
+      # allowVolumeExpansion: true
+      # # The Cinder availability zone to use for volumes provisioned by the storage class
+      # availabilityZone: nova
+      # # The Cinder volume type to use for volumes provisioned by the storage class
+      # # If not given, the default volume type will be used
+      # volumeType: fast
+      # # The allowed topologies for the storage class
+      # allowedTopologies:
```
mkjpryor, line 103: Please keep the same defaults as we had previously so as not to break existing installations that just use the defaults.

charts/cluster-addons/values.yaml
```diff
@@ -89,25 +88,32 @@ openstack:
...
-    storageClass:
-      # Indicates if the storage class should be enabled
-      enabled: true
-      # The name of the storage class
-      name: csi-cinder
-      # Indicates if the storage class should be annotated as the default storage class
-      isDefault: true
-      # The reclaim policy for the storage class
+    defaultStorageClass:
+      name: csi-cinder-fast
```
mkjpryor, line 92: This should be just `csi-cinder` to match the old default

charts/cluster-addons/values.yaml
```diff
@@ -89,25 +88,32 @@ openstack:
...
-      volumeType:
-      # The allowed topologies for the storage class
+      # # The Cinder volume type to use for volumes provisioned by the storage class
+      # # If not given, the default volume type will be used
+      volumeType: fast
+      # # The allowed topologies for the storage class
       allowedTopologies:
+    additionalStorageClasses: []
+      # - name: csi-cinder-fast
+      #   # The reclaim policy for the storage class
```
mkjpryor, line 106: This should document `volumeBindingMode` as being required for additional storage classes

mkjpryor (changes requested):
Sorry @heytrav - just a few more things that I missed the first time

charts/cluster-addons/values.yaml
```diff
@@ -8,7 +7,7 @@ clusterName:
 hooks:
   image:
     repository: ghcr.io/stackhpc/capi-helm-utils
-    tag:  # Defaults to chart appVersion if not given
+    tag: # Defaults to chart appVersion if not given
```
mkjpryor, line 10: This space is important for the linting

mkjpryor (changes requested):

mkjpryor:
@heytrav Some changes were made to the cloud where we run the CI. Please can you rebase to pick up the required config changes?

mkjpryor (approved):
LGTM

mkjpryor:
@heytrav 

The chart upgrade test is failing with the following error:

```
Error: UPGRADE FAILED: cannot patch "csi-cinder" with kind StorageClass: StorageClass.storage.k8s.io "csi-cinder" is invalid: parameters: Forbidden: updates to parameters are forbidden.
```

Looking at the patch, I think this is due to attempting to add `fstype: ext4` to the storage class. To be consistent with the previous setup, we need to not set this for the default storage class. I think the default is `ext4` anyway?

mkjpryor:
I feel like using the Kubernetes default fstype by default is the desired behaviour anyway, rather than the chart having an explicit opinion about it being `ext4`.

charts/cluster-addons/values.yaml
```diff
@@ -84,8 +79,30 @@ openstack:
       # The Cinder volume type to use for volumes provisioned by the storage class
       # If not given, the default volume type will be used
       volumeType:
+      # The volumeBindingMode field controls when volume binding and dynamic provisioning should occur.
+      volumeBindingMode: WaitForFirstConsumer
       # The allowed topologies for the storage class
       allowedTopologies:
+      # Filesystem type supported by Kubernetes. Defaults to "ext4"
+      fstype: ext4
```
mkjpryor, line 87: Do not set a default here.

mkjpryor (changes requested):

charts/cluster-addons/values.yaml
heytrav: Ok I removed ext4 as the default

mkjpryor:
@heytrav Did you mean to do all that other reformatting of the values file?

heytrav:
> @heytrav Did you mean to do all that other reformatting of the values file?

I didn't mean to no. My editor's linter settings are a bit too opinionated.

mkjpryor (approved):
LGTM subject to CI

PR #177, opened 2023-12-13
Remove unused image build

mkjpryor:


mkjpryor:
Merging without passing CI as we use `pull_request_target`.

PR #223, opened 2024-01-19
Ability to put etcd on a separate block device

mkjpryor:


scrungus (approved):
LGTM

scrungus:
CI failure seems unrelated to this at least - kube-upgrade which included etc test passed

mkjpryor:
The CI failure is because there is only one spare router in the test tenancy which was being consumed by another CI run. Passed on a re-run.

PR #230, opened 2024-01-25
Use less aggressive node startup timeouts

mkjpryor:
This should reduce node churn in the case of a transient issue, e.g. a network problem prevents access to OpenStack APIs.

### DIFF
--- a/charts/openstack-cluster/values.yaml
+++ b/charts/openstack-cluster/values.yaml
@@ -292,8 +292,8 @@ controlPlane:
     spec:
       # By default, unhealthy control plane nodes are always remediated
       maxUnhealthy: 100%
-      # If a node takes longer than 10 mins to startup, remediate it
-      nodeStartupTimeout: 10m0s
+      # If a node takes longer than 30 mins to startup, remediate it
+      nodeStartupTimeout: 30m0s
       # By default, consider a control plane node that has not been Ready
       # for more than 5 mins unhealthy
       unhealthyConditions:
@@ -389,8 +389,8 @@ nodeGroupDefaults:
     spec:
       # By default, unhealthy worker nodes are always remediated
       maxUnhealthy: 100%
-      # If a node takes longer than 10 mins to startup, remediate it
-      nodeStartupTimeout: 10m0s
+      # If a node takes longer than 30 mins to startup, remediate it
+      nodeStartupTimeout: 30m0s
       # By default, consider a worker node that has not been Ready for
       # more than 5 mins unhealthy
       unhealthyConditions:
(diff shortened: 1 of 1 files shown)


charts/openstack-cluster/values.yaml
```diff
@@ -378,8 +378,8 @@ nodeGroupDefaults:
     spec:
       # By default, unhealthy worker nodes are always remediated
       maxUnhealthy: 100%
```
JohnGarbutt, line 380: # (Optional) maxUnhealthy prevents further remediation if the cluster is already partially unhealthy
  maxUnhealthy: 40%

Maybe 100% is the wrong call here. I think its actually a limit around remediation.

charts/openstack-cluster/values.yaml
```diff
@@ -281,8 +281,8 @@ controlPlane:
     spec:
       # By default, unhealthy control plane nodes are always remediated
       maxUnhealthy: 100%
-      # If a node takes longer than 10 mins to startup, remediate it
-      nodeStartupTimeout: 10m0s
+      # If a node takes longer than 30 mins to startup, remediate it
+      nodeStartupTimeout: 30m0s
```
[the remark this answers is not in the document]
JohnGarbutt: I am specifically thinking about allows for the reboot of a physical server here, which might take 15-20 mins.

charts/openstack-cluster/values.yaml
JohnGarbutt: I think we might want to set this number to 1, or something like that, so it limits the number of unhealth nodes we rebuild at any one time?
https://cluster-api.sigs.k8s.io/tasks/automated-machine-management/healthchecking#max-unhealthy

JohnGarbutt (changes requested):
So I think for worker nodes, we probably should set maxUnhealthy to 1 by default, i.e. only take out one node at once when trying to remediate.

We could probably do the same for the control plane I guess, although I think you have a different plan there already?

Also, let's take a look at that unhealthyConditions? A baremetal node might take 15-20 mins to reboot. Should we leave time for that to happen? If so which of those would we tweak? I guess we can test a VM reboot to see if its similar?

charts/openstack-cluster/values.yaml
```diff
@@ -290,10 +290,10 @@ controlPlane:
     enabled: true
     # The spec for the health check
     spec:
-      # By default, unhealthy control plane nodes are always remediated
-      maxUnhealthy: 100%
-      # If a node takes longer than 10 mins to startup, remediate it
-      nodeStartupTimeout: 10m0s
+      # By default, don't remediate control plane nodes when more than one is unhealthy
+      maxUnhealthy: 1
```
JohnGarbutt, line 294: Maybe lets do 40% here as well?

JohnGarbutt: ... just trying to convince myself that this will allow one node to be fixed, I guess it does.

JohnGarbutt (approved):
Ok, lets try this out.

PR #251, opened 2024-02-11
Add global network policy to block access to metaservice

heytrav:
* Use calico GlobalNetworkPolicy to deny egress to metaservice
* basic values to for priority, deny and allow CIDR

### DIFF
--- a/charts/cluster-addons/templates/cni/calico.yaml
+++ b/charts/cluster-addons/templates/cni/calico.yaml
@@ -49,6 +49,73 @@ spec:
     - secret:
         name: {{ include "cluster-addons.componentName" (list . "cni-calico") }}-config
         key: overrides
+{{- if .Values.cni.calico.globalNetworkPolicy }}
+---
+apiVersion: addons.stackhpc.com/v1alpha1
+kind: Manifests
+metadata:
+  name: {{ include "cluster-addons.componentName" (list . "cni-calico-globalnetpolicy") }}
+  labels: {{ include "cluster-addons.componentLabels" (list . "cni-calico-globalnetpolicy") | nindent 4 }}
+  annotations:
+    # Tell Argo to ignore the non-controller owner references for this object
+    argocd.argoproj.io/sync-options: "ControllerReferencesOnly=true"
+spec:
+  clusterName: {{ include "cluster-addons.clusterName" . }}
+  bootstrap: true
+  targetNamespace: {{ .Values.cni.calico.release.namespace }}
+  releaseName: cni-calico-metadata-global-netpolicy
+  manifestSources:
+    - template: |
+        ---
+        apiVersion: projectcalico.org/v3
+        kind: GlobalNetworkPolicy
+        metadata:
+          name: {{ include "cluster-addons.componentName" (list . "deny-egress") }}
+        spec:
+          order: {{ .Values.cni.calico.globalNetworkPolicy.denyPriority }}
+          namespaceSelector: {{ .Values.cni.calico.globalNetworkPolicy.denyNamespaceSelector }}
+          types:
+            - Egress
+          egress:
+            - action: Deny
+              ipVersion: 4
+              destination:
+                nets:
+                {{- range .Values.cni.calico.globalNetworkPolicy.denyEgressCidrs }}
+                  - {{ . }}
+                {{ end }}
+            - action: Deny
+              ipVersion: 6
+              destination:
+                nets:
+                {{- range .Values.cni.calico.globalNetworkPolicy.denyv6EgressCidrs }}
+                  - {{ . }}
+                {{ end }}
+        ---
+        apiVersion: projectcalico.org/v3
+        kind: GlobalNetworkPolicy
+        metadata:
+          name: {{ include "cluster-addons.componentName" (list . "allow-global-egress") }}
+        spec:
+          order: {{ .Values.cni.calico.globalNetworkPolicy.allowPriority }}
+          types:
+            - Egress
+          egress:
+            - action: Allow
+              ipVersion: 4
+              destination:
+                nets: 
+                {{- range .Values.cni.calico.globalNetworkPolicy.allowEgressCidrs }}
+                  - {{ . }}
+                {{ end }}
+            - action: Allow
+              ipVersion: 6
+              destination:
+                nets: 
+                {{- range .Values.cni.calico.globalNetworkPolicy.allowv6EgressCidrs }}
+                  - {{ . }}
+                {{ end }}
+{{- end }}
 {{- if .Values.monitoring.enabled }}
 ---
 apiVersion: addons.stackhpc.com/v1alpha1
--- a/charts/cluster-addons/values.yaml
+++ b/charts/cluster-addons/values.yaml
@@ -20,6 +20,20 @@ cni:
     release:
       namespace: tigera-operator
       values: {}
+    # Nova metadata service
+    # See https://docs.openstack.org/nova/latest/user/metadata.html#the-metadata-service
+    globalNetworkPolicy:
+      denyNamespaceSelector: kubernetes.io/metadata.name != 'openstack-system'
+      allowPriority: 20
+      denyPriority: 10
+      allowEgressCidrs:
+        - "0.0.0.0/0"
+      denyEgressCidrs:
+        - "169.254.169.254/32"
+      allowv6EgressCidrs:
+        - "::/0"
+      denyv6EgressCidrs:
+        - "fe80::a9fe:a9fe/128"
   # Settings for the Cilium CNI
   # See https://docs.cilium.io/en/stable/gettingstarted/k8s-install-helm/ for details
   cilium:


mkjpryor:
Doesn't the OCCM need to access the metadata service?

heytrav:
> Doesn't the OCCM need to access the metadata service?

Good point! `cinder-csi-controllerplugin` does as well I believe. I'm a little torn on whether to open it for the whole namespace or just specific pods. 

I'll just change it to allow the `openstack-system` namespace for now.

charts/cluster-addons/templates/cni/calico.yaml
```diff
@@ -49,6 +49,92 @@ spec:
...
+        spec:
+          order: {{ .Values.cni.calico.metadataGlobalNetworkPolicy.metadataDenyPriority }}
+          types:
+            - Egress
+          egress:
+            - action: Deny
+              destination:
+                nets:
+                  {{- range .Values.cni.calico.metadataGlobalNetworkPolicy.denyEgressCidrs }}
+                    - {{ . }}
```
mkjpryor, line 84: Reduce indent by two here (just for style) and other similar places

heytrav:
> More generally, I'm not sure why we need three policies instead of a single policy with `namespaceSelector: namespaceSelector: kubernetes.io/metadata.name != 'openstack-system'` that denies egress only to the metadata service?
> 
> Shouldn't this be sufficient?
> 
> ```yaml
> apiVersion: projectcalico.org/v3
> kind: GlobalNetworkPolicy
> metadata:
>   name: {{ include "cluster-addons.componentName" (list . "deny-metadata") }}
> spec:
>   order: {{ .Values.cni.calico.metadataGlobalNetworkPolicy.priority }}
>   namespaceSelector: kubernetes.io/metadata.name != 'openstack-system'
>   types:
>     - Egress
>   egress:
>     - action: Deny
>       ipVersion: 4
>       destination:
>         nets: 
>           - 169.254.169.254/32
>     - action: Deny
>       ipVersion: 6
>       destination:
>         nets: 
>           - fe80::a9fe:a9fe/128
> ```

I also would have thought that this would work, and I did try this originally. However when I do that then _nothing_ works except for not-blocking egress to metadata in the target namespace.  Here is  what I tried:

### No GlobalNetworkPolicy


```

# kubectl exec into pod
# test-policy5 pod in namespace "test-policy"
root@test-policy5:/# curl 169.254.169.254
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04
latestroot@test-policy5:/# exit

# kubectl exec into 
# test-policy4 pod in namespace "default"


root@test-policy4:/# curl 169.254.169.254
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04
latestroot@test-policy4:/# curl https://google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="https://www.google.com/">here</A>.
</BODY></HTML>



```


### Add following GlobalNetworkPolicy


```

---
apiVersion: projectcalico.org/v3
kind: GlobalNetworkPolicy
metadata:
  name: calico-test-deny-metaservice-egress
spec:
  namespaceSelector: kubernetes.io/metadata.name != 'test-policy'
  order: 10
  types:
    - Egress
  egress:
    - action: Deny
      ipVersion: 4
      destination:
        nets:
          - 169.254.169.254/32

    - action: Deny
      ipVersion: 6
      destination:
        nets:
          - fe80::a9fe:a9fe/128

```

# Test pod in `test-policy` namespace

```

root@test-policy5:/# curl 169.254.169.254
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04


latestroot@test-policy5:/# curl https://google.com
curl: (6) Could not resolve host: google.com


root@test-policy5:/# exit

```
### Test pod in `default` namespace
```
# In pod test-policy4 in namespace "default"

root@test-policy4:/# curl 169.254.169.254 # hangs for several minutes
^C
root@test-policy4:/# curl https://google.com
curl: (6) Could not resolve host: google.com
root@test-policy4:/# 


```

So with just the one GlobalNetworkPolicy it  is denying egress to everything except the metadata service in the `test-policy` namespace.

If I add the global egress allow with lower priority as well as the deny policy it seems to work ok:

```
---
apiVersion: projectcalico.org/v3
kind: GlobalNetworkPolicy
metadata:
  name: calico-test-deny-metaservice-egress
spec:
  namespaceSelector: kubernetes.io/metadata.name != 'test-policy'
  order: 10
  types:
    - Egress
  egress:
    - action: Deny
      ipVersion: 4
      destination:
        nets:
          - 169.254.169.254/32

    - action: Deny
      ipVersion: 6
      destination:
        nets:
          - fe80::a9fe:a9fe/128
---
apiVersion: projectcalico.org/v3
kind: GlobalNetworkPolicy
metadata:
  name: calico-test-allow-global-egress
spec:
  order: 20
  types:
    - Egress
  egress:
    - action: Allow
      ipVersion: 4
      destination:
        nets:
          - 0.0.0.0/0

    - action: Allow
      ipVersion: 6
      destination:
        nets:
          - ::/0

```

### In pod in `test-policy` namespace

```
root@test-policy5:/# curl 169.254.169.254
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04


latestroot@test-policy5:/# curl https://google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="https://www.google.com/">here</A>.
</BODY></HTML>
root@test-policy5:/# 
```

### in `default` namespace

```

root@test-policy4:/# curl 169.254.169.254
^C


root@test-policy4:/# curl https://google.com
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="https://www.google.com/">here</A>.
</BODY></HTML>
root@test-policy4:/# 

```

mkjpryor:
Maybe having **a** policy makes it default to Deny 🤷‍♂️

charts/cluster-addons/templates/cni/calico.yaml
heytrav: I moved some lines over. I think  you meant the list below  `nets:` right?

mkjpryor:
It seems like this change is actually just a mechanism for setting global allow/deny CIDRs. Maybe we should remove metadata from the name of the values and resources?

heytrav:
> It seems like this change is actually just a mechanism for setting global allow/deny CIDRs. Maybe we should remove metadata from the name of the values and resources?

makes sense. done

charts/cluster-addons/templates/cni/calico.yaml
```diff
@@ -49,6 +49,73 @@ spec:
...
+  manifestSources:
+    - template: |
+        ---
+        apiVersion: projectcalico.org/v3
+        kind: GlobalNetworkPolicy
+        metadata:
+          name: {{ include "cluster-addons.componentName" (list . "deny-egress") }}
+        spec:
+          order: {{ .Values.cni.calico.globalNetworkPolicy.denyPriority }}
+          namespaceSelector: kubernetes.io/metadata.name != 'openstack-system'
```
mkjpryor, line 76: Maybe we should make this a Helm value as well, as it might depend on what the user sets for `denyEgressCidrs`?

mkjpryor (changes requested):

charts/cluster-addons/templates/cni/calico.yaml
```diff
@@ -49,6 +49,73 @@ spec:
...
+  manifestSources:
+    - template: |
+        ---
+        apiVersion: projectcalico.org/v3
+        kind: GlobalNetworkPolicy
+        metadata:
+          name: {{ include "cluster-addons.componentName" (list . "deny-egress") }}
+        spec:
+          order: {{ .Values.cni.calico.globalNetworkPolicy.denyPriority }}
+          namespaceSelector: kubernetes.io/metadata.name != '{{ .Values.cni.calico.globalNetworkPolicy.allowNamespace }}'
```
mkjpryor, line 76: Ah - I was thinking the whole selector becomes a Helm value. Presumably there are filters other than `!=` that you can use.

mkjpryor (changes requested):

charts/cluster-addons/templates/cni/calico.yaml
mkjpryor: Sorry that I wasn't clear the first time 😃

mkjpryor (approved):
LGTM

PR #252, opened 2024-02-13
Fix metrics retention size constraint

mkjpryor:
Also tweak default log retention

### DIFF
--- a/charts/cluster-addons/values.yaml
+++ b/charts/cluster-addons/values.yaml
@@ -223,7 +223,7 @@ monitoring:
             compactor:
               retention_enabled: true
             limits_config:
-              retention_period: 744h
+              retention_period: 24h
           # Enable persistence by default
           persistence:
             enabled: true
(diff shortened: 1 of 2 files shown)


charts/cluster-addons/values.yaml
```diff
@@ -223,7 +223,7 @@ monitoring:
             compactor:
               retention_enabled: true
             limits_config:
-              retention_period: 744h
+              retention_period: 24h
```
MaxBed4d, line 226: I think that 24h may be a little low, especially if this is for how long logs should be kept for, as it won't give users the chance to spot any abnormalities that may occur over, let's say, a weekend.

JohnGarbutt: Our take was, things go really bad when this overflows, and most logs keep happening again and again.

However, ensuring you can check what happened Friday at 5pm on a Monday morning is a very good point. I would be OK trying for 72h instead here.

MaxBed4d (commented):
Looks good to go!

PR #253, opened 2024-02-13
Add pod disruption budget to metrics-server

mkjpryor:
Pods for the metrics-server are currently in `kube-system` without a pod disruption budget, which is a category of pods that prevent scale down in auto-scaling clusters.


MaxBed4d (approved):
Looks good to me!

JohnGarbutt (approved):
ah, I see, this seems correct for this service.

PR #254, opened 2024-02-13
Improve cluster autoscaler default config

mkjpryor:
In particular, allow pods with `emptyDir`s to be evicted, switch to the `least-waste` expander and increase the default log level.


m-bull (approved):
Looks good - thanks for the fix

JohnGarbutt (approved):
Ah, yes, this looks pretty vital given the conversation earlier

charts/openstack-cluster/templates/autoscaler/deployment.yaml
MaxBed4d, line 1: Looks good to me - thank you for fixing this 🫡

charts/openstack-cluster/templates/autoscaler/role.yaml
MaxBed4d, line 1: Looks good, again, thank you!

MaxBed4d (approved):
All looks good, I had just that one question, but apart from that I am happy with this.

PR #262, opened 2024-02-18
Reduce indentation of allowedCidrs

heytrav:
* allowedCidrs is separate from loadBalancerProvider but was being rendered as child of that key.


mkjpryor (approved):
LGTM

PR #280, opened 2024-03-07
Disable volume topology constraints by default

mkjpryor:


JohnGarbutt (approved):
This makes sense. Got confused reading around, particularly why the old workaround has got broken: `ignore-volume-az: true` But that is a distraction, this seems like a step forward.

PR #297, opened 2024-04-04
Fix log ordering in Loki dashboards

sd109:
At the moment these dashboards use newest first ordering which means that most recent log lines appear at the top rather than the bottom. This makes certain things difficult to read e.g. Python stack traces get split by lines so the trace lines appear reversed in the logs. This change fixes the ordering to make reading the logs easier.


mkjpryor:
As discussed elsewhere, I don't think this is really the solution. Keeping the most recent logs at the top makes sense.

Instead, we should look at moving components to use structured logging as much as possible.

PR #303, opened 2024-04-10
Update k8s-keystone-auth subchart with working image

heytrav:
Just discovered that the image referenced in helm chart doesn't exist. Updated with a version that has an existing image.


mkjpryor (approved):
LGTM

PR #337, opened 2024-05-15
Add workaround for pod instabilities observed when using Nvidia's container runtime

sd109:
We're seeing intermittent issues with NGINX controller and Alertmanager (and CoreDNS) pods when they are scheduled to nodes which are using the Nvidia container runtime. The root cause of the issue needs more investigation, but adding some affinity rules where possible to discourage the problematic pods from landing on GPU nodes will improve the situation in the short term.

### DIFF
--- a/charts/cluster-addons/values.yaml
+++ b/charts/cluster-addons/values.yaml
@@ -183,7 +183,21 @@ ingress:
       version: 4.10.1
     release:
       namespace: ingress-nginx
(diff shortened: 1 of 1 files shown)


mkjpryor:
I understand why you want this patch, but I really hate it. We need to get to the bottom of the issues really. I'd prefer this was maintained in site-level Azimuth configs TBH.

PR #356, opened 2024-06-25
Add node problem detector as a default addon

mkjpryor:


scrungus:
Looks good. How does this tie in with the monitoring stack - does it just export a prometheus metric? Can it connect to Loki?

PR #366, opened 2024-07-12
Increase worker node maxUnhealthy remediation threshold to 100%

sd109:
As discussed elsewhere, we now have feature enabled which blocks all worker node remediation if control plane is unhealthy, so should be safe to reset worker remediation threshold to 100% without the risk of runaway remediations seen previously.


mkjpryor (approved):
LGTM

PR #369, opened 2024-07-29
point to new occm

scrungus:


JohnGarbutt:
@scrungus are we OK to drop this PR now? I guess this has all merged upstream now, has it been released and merged into our repo yet?

scrungus:
yep, we can drop this

PR #370, opened 2024-07-30
Do not enable optional addons

waipeng:
The Nvidia and Mellanox addons are not necessary for a cluster; do not enable them for now.


waipeng:
Should we do it here or in `charts/openstack-cluster/values.yaml`? Looking for advice on consistency. E.g. 

For `openstack` addon, it is 
cluster-addons - `enabled: false`
openstack-cluster - `enabled: true`

For `nvidiaGPUOperator` addon, it is
cluster-addons - `enabled: true`
openstack-cluster - not defined

pawcykca:
I think the best place would be the values ​​object generated by the Magnum CAPI Helm driver. This way we will avoid changes coming from this or another repository providing Helm Charts for Magnum CAPI Helm driver.

PR #388, opened 2024-08-02
Allow configuration of Calico bgp and encapsulation values

dalees:
To enable Calico BGP mode we need to set bgp and encapsulation values. The chart allows overriding of the bgp value already, but encapsulation is within a list, which is concatenated if changes are attempted in the overrides field.

So, here we add configuration for both values and document how to enable the full-mesh mode we will use.

### DIFF
--- a/charts/cluster-addons/templates/cni/calico.yaml
+++ b/charts/cluster-addons/templates/cni/calico.yaml
@@ -13,7 +13,7 @@ stringData:
       nodeMetricsPort: 9091
       typhaMetricsPort: 9093
       calicoNetwork:
-        bgp: Disabled
+        bgp: {{ .Values.cni.calico.bgp }}
         nodeAddressAutodetectionV4:
           kubernetes: NodeInternalIP
         ipPools:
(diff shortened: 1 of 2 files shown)


dalees:
Note: This should result in no default behaviour change. An implementation of using these options will be submitted to the `magnum-capi-helm` driver.

dalees:
Created Magnum CAPI driver change that depends on this PR: https://review.opendev.org/c/openstack/magnum-capi-helm/+/925552

mkjpryor (changes requested):
I'm not overly keen on essentially moving parts of the Helm values for the chart out of the `values` section, as it makes us more brittle against changes to the chart values.

In particular the new `calico.bgp` setting proposed here is essentially just a shorthand for `calico.release.values.installation.calicoNetwork.bgp`.

There are more complications with the encapsulation, but I think maybe we can accomplish the same effect by inspecting the values that are given and adapting the defaults accordingly, as happens in some other addons (e.g. only generating the default IP pool if one is not given in the values).

PR #390, opened 2024-08-13
Truncate secret names to prevent going over 63 char limit.

dalees:
Some secrets created for very long cluster names were going over the 63 char label limit of Kubernetes, and were not deploying.

Example error message:
  failureMessage: 'HelmRelease.addons.stackhpc.com "reallylongnamecluster1reallylo-oe62woigm3qq-ccm-openstack"
    is invalid: metadata.labels: Invalid value: "secret.addons.stackhpc.com/reallylongnamecluster1reallylo-oe62woigm3qq-ccm-openstack-config":
    name part must be no more than 63 characters'

### DIFF
--- a/charts/cluster-addons/templates/_helpers.tpl
+++ b/charts/cluster-addons/templates/_helpers.tpl
@@ -7,11 +7,12 @@ The name of the target cluster.
 
 {{/*
 Create a name for a cluster component.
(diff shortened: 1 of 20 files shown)


mkjpryor:
@dalees 

We also noticed this, and fixed it in the addon provider instead:

https://github.com/azimuth-cloud/cluster-api-addon-provider/pull/115

Issue #398, opened 2024-09-02
API Provider mismatch issues when using cluster api provider 0.10.4  


anish-mudaraddi:
I've created a management cluster using these charts (using clusterctl openstack provider version 0.10.4). 
I'm using ArgoCD for managing deployments. 

I'm getting out-of-sync issues for these lines: 
- https://github.com/azimuth-cloud/capi-helm-charts/blob/main/charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml#L94
- https://github.com/azimuth-cloud/capi-helm-charts/blob/main/charts/openstack-cluster/templates/node-group/machine-deployment.yaml#L65
- https://github.com/azimuth-cloud/capi-helm-charts/blob/main/charts/openstack-cluster/templates/cluster.yaml#L16

Where the current `infrastructureRef.apiVersion` is currently `infrastructure.cluster.x-k8s.io/v1beta1` but wants to be `apiVersion: infrastructure.cluster.x-k8s.io/v1alpha7` 

Is there any reason why we can't change all of these `APIVersion` references - https://github.com/search?q=repo%3Aazimuth-cloud%2Fcapi-helm-charts%20v1alpha7&type=code to `infrastructure.cluster.x-k8s.io/v1beta1` since this version was made available in v0.10.4?

anish-mudaraddi:
thanks @mkjpryor for fixing this in the latest chart version!

PR #399, opened 2024-09-02
update API Provider to v1beta1

anish-mudaraddi:
An attempt to fix issue https://github.com/azimuth-cloud/capi-helm-charts/issues/398

update api provider to use v1beta1 since its provided in newest cluster api provider version 0.10.*

### DIFF
--- a/charts/openstack-cluster/templates/cluster-openstack.yaml
+++ b/charts/openstack-cluster/templates/cluster-openstack.yaml
@@ -25,9 +25,15 @@ spec:
     {{- toYaml .manageSecurityGroups | nindent 2 }}
   {{- end }}
   {{- end }}
-  {{- with .externalNetworkId }}
+  disableExternalNetwork: {{ .disableExternalNetwork | default "false" }}
+  {{- if or .externalNetworkFilter .externalNetworkId }}
   externalNetwork:
-    id: {{ . }}
+    {{- if .externalNetworkFilter }}
+    {{- toYaml .externalNetworkFilter | nindent 2 }}
+    {{- else }}
+    id: {{ .externalNetworkId }}
+  {{- end }}
+  {{- end }}
   {{- end }}
   {{- with .internalNetwork }}
   {{- if or .networkFilter .subnetFilter }}
--- a/charts/openstack-cluster/templates/cluster-openstack.yaml
+++ b/charts/openstack-cluster/templates/cluster-openstack.yaml
@@ -11,20 +11,19 @@ metadata:
     # NOTE: Argo won't delete this object itself as it has an owner reference to the cluster
 spec:
   identityRef:
-    kind: Secret
     name: {{ include "openstack-cluster.cloudCredentialsSecretName" . }}
-  cloudName: openstack
+    cloudName: openstack
   {{- with .Values.controlPlaneEndpoint }}
   controlPlaneEndpoint: {{ toYaml . | nindent 4 }}
   {{- end }}
   {{- with .Values.clusterNetworking }}
-  {{- with .dnsNameservers }}
-  dnsNameservers: {{ toYaml . | nindent 4 }}
+  {{- if .manageSecurityGroups }}
+  managedSecurityGroups: 
+    allowAllInClusterTraffic: {{ .allowAllInClusterTraffic }}
   {{- end }}
-  managedSecurityGroups: {{ .manageSecurityGroups }}
-  allowAllInClusterTraffic: {{ .allowAllInClusterTraffic }}
   {{- with .externalNetworkId }}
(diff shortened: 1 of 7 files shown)


scrungus:
I think this will require some additional changes, for example removing `identityRef.Kind`:
https://github.com/azimuth-cloud/capi-helm-charts/blob/e4e1ec193d02dc44a19a8cf9f3b4fd5656a7cd5f/charts/openstack-cluster/templates/node-group/openstack-machine-template.yaml#L11

Could you go through [this](https://cluster-api-openstack.sigs.k8s.io/topics/crd-changes/v1alpha7-to-v1beta1#removal-of-machine-identityrefkind) changelog and make any necessary changes?

scrungus (changes requested):
I think this will require some additional changes, for example removing `identityRef.kind`:
https://github.com/azimuth-cloud/capi-helm-charts/blob/e4e1ec193d02dc44a19a8cf9f3b4fd5656a7cd5f/charts/openstack-cluster/templates/node-group/openstack-machine-template.yaml#L11
https://github.com/azimuth-cloud/capi-helm-charts/blob/e4e1ec193d02dc44a19a8cf9f3b4fd5656a7cd5f/charts/openstack-cluster/templates/cluster-openstack.yaml#L14
and addition of `cloudName`. 

Could you go through [this](https://cluster-api-openstack.sigs.k8s.io/topics/crd-changes/v1alpha7-to-v1beta1#removal-of-machine-identityrefkind) changelog and make any necessary changes?

anish-mudaraddi:
I've made the necessary changes - and I've tested making a cluster with the changes and its working for me!

scrungus:
@anish-mudaraddi the etcd test is failing, looks like we need [this](https://cluster-api-openstack.sigs.k8s.io/topics/crd-changes/v1alpha7-to-v1beta1#change-to-availabilityzone-for-additionalblockdevices) change as well.

mkjpryor (changes requested):
This PR is a good start, but while we are moving to a new API version I'd like to take the opportunity to adopt some of the new functionality. Some examples would be:

* Moving the external network specification to a proper filter but still respecting the old-style ID for backwards compatibility
* Supporting multiple subnets but still respecting the old-style single subnet configuration for backwards compatibility
* ...

I understand that you want this quickly to correct your issues with Argo CD, but it seems like a good opportunity to also do these things.

mkjpryor:
It is also important that all of these changes are implemented in a backwards-compatible way. This means that where fields have changed their name, e.g. `diskSize` to `sizeGiB` for root volumes, we want to respect both.

anish-mudaraddi:
@mkjpryor no problem - I'll try to implement the functionality you requested!

anish-mudaraddi:
I've implemented the following changes:
- removal of identityRef kind Secret
- moving `cloudName` to `identityRef.cloudName`
- added field `ServerGroupFilter` - (an alias for `ServerGroup`)
    - `ServerGroupID` is still supported for backwards compatibility. 
    - I used `ServerGroupFilter` instead of `ServerGroup` to match existing naming convention used for defining subnets and networks - `subnetFilter` and `networkFilter` 
    
- added field `externalNetworkFilter` - (similar to above)
- added fields `managedSubnets` and `subnets` to match upstream 
    - old fields `nodeCidr` and `dnsNameservers` still supported 
 
- add field `sizeGiB` to replace `diskSize`  
    - supporting both for backwards compatibility

- added field `allowedCIDRs` and `allowedCidrs`

- changes to the way `machineImage` and `machineImageId` are parsed to be compatible with new changes  

- availabilityZone field in v1beta1 has changed to no longer accept name string
    - I updated this but also support defining the name the old way

- added `disableExternalNetwork` field    
    
**Bits I left out**

- a field like `machineimageFilter` - I don't think we need it since we can define an image via id and name currently which I think is enough

- `floatingIPPoolRef` functionality - something for a future PR - I think its out-of-scope for this PR

anish-mudaraddi:
@mkjpryor @scrungus  just bumping this

mkjpryor:
@anish-mudaraddi Can you rebase this onto `main` please? Thanks!

mkjpryor:
@anish-mudaraddi

I'm finding this PR very hard to review.

Thinking on it some more, I think I would prefer to split this change into two PRs:

  1. Move to `v1beta1` resources with no changes to the `values.yaml`
  2. Introduce some of the new features, in a backwards compatible way

mkjpryor:
@anish-mudaraddi 

I think #423 is the minimal set of changes needed to move to `v1beta1` resources while supporting the existing values format. Can you please test and comment on that PR?

mkjpryor:
Superseded by #423 for now

PR #419, opened 2024-10-03
Add some debugging docs

### LINKED ISSUE #232: Unit tests and debug tools
When these charts work, they are fine. When they don't work, the user is left in a world of hurt, rummaging round in a complex system trying to find clues as to what has gone wrong.

For a production system intended for others to use, these charts should come with a set of unit tests and debug tools that can be run on a cluster to check all the components are present and correct. 

A simple library of `kubectl` commands to look for obvious things would be a useful start. So rather than posting suggestions in a Slack channel, StackHPC can point people at a page of common debug queries they can copy and paste.

These scripts should include `kubectl` queries to resolve the random Pod names so that they will _'just work'_ without requiring the user to resolve the names manually.

These will be useful for people trying to port the Helm charts to new platforms, and for system admins trying to verify if their Openstack platform is working correctly.

sd109:
Closes #232

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,3 @@ Currently, the following charts are available:
 | [cluster-addons](./charts/cluster-addons) | Deploys addons into a Kubernetes cluster, e.g. CNI. |
 | [etcd-defrag](./charts/etcd-defrag/) | Installs a `CronJob` for running [etcd defragmentation](https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation). |
 | [openstack-cluster](./charts/openstack-cluster) | Deploys a Kubernetes cluster on an OpenStack cloud. |
-
-# Usage
-
-## OpenStack
-
-See the openstack-cluster chart [README.md](./charts/openstack-cluster/README.md) for deployment instructions and [DEBUGGING.md](./charts/openstack-cluster/DEBUGGING.md) for debugging tips.
--- a/README.md
+++ b/README.md
@@ -25,3 +25,9 @@ Currently, the following charts are available:
 | [cluster-addons](./charts/cluster-addons) | Deploys addons into a Kubernetes cluster, e.g. CNI. |
 | [etcd-defrag](./charts/etcd-defrag/) | Installs a `CronJob` for running [etcd defragmentation](https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation). |
 | [openstack-cluster](./charts/openstack-cluster) | Deploys a Kubernetes cluster on an OpenStack cloud. |
+
+# Usage
+
+## OpenStack
+
+See the openstack-cluster chart [README.md](./charts/openstack-cluster/README.md) for deployment instructions and [DEBUGGING.md](./charts/openstack-cluster/DEBUGGING.md) for debugging tips.
--- a/charts/openstack-cluster/DEBUGGING.md
+++ b/charts/openstack-cluster/DEBUGGING.md
@@ -1,0 +1,174 @@
+# Debugging
+
+## Cluster Creation Summary
+
+The following diagram represents a high-level summary of cluster creation process. Each stage is described in more detail below.
+
+```mermaid
+sequenceDiagram
+  actor U as User
+  participant M as Management Cluster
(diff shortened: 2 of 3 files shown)


charts/openstack-cluster/DEBUGGING.md
```diff
@@ -0,0 +1,172 @@
...
+  W -->> M: Wait for workload cluster API server to be online
+  M ->> W: Install addons (helmreleases & manifests)
+  activate W
+  W -->> M: Wait for core addons (e.g. CNI) to be ready
+  M ->> O: Create remaining control plane and worker nodes
+  O ->> W: Create remaining servers
+  W -->> M: Wait for all addons to be ready
+  deactivate W
+  M -) U: Cluster ready
+```
```
sd109, line 30: I'm still on the fence about whether this diagram is really useful here, happy to remove it depending on others' opinions.

README.md
```diff
@@ -25,3 +25,9 @@ Currently, the following charts are available:
 | [cluster-addons](./charts/cluster-addons) | Deploys addons into a Kubernetes cluster, e.g. CNI. |
 | [etcd-defrag](./charts/etcd-defrag/) | Installs a `CronJob` for running [etcd defragmentation](https://etcd.io/docs/v3.5/op-guide/maintenance/#defragmentation). |
 | [openstack-cluster](./charts/openstack-cluster) | Deploys a Kubernetes cluster on an OpenStack cloud. |
+
+# Usage
+
+## OpenStack
+
+See the openstack-cluster chart [README.md](./charts/openstack-cluster/README.md) for deployment instructions and [DEBUGGING.md](./charts/openstack-cluster/DEBUGGING.md) for debugging tips.
```
mkjpryor, line 33: Don't know if this is required? I think linking the debugging page from the `openstack-cluster` README would be better?

charts/openstack-cluster/DEBUGGING.md
```diff
@@ -0,0 +1,172 @@
+# Debugging
```
mkjpryor, line 1: Maybe "Debugging a Cluster API cluster on OpenStack"?

charts/openstack-cluster/DEBUGGING.md
```diff
@@ -0,0 +1,172 @@
...
+
+Commonly encountered issues (usually visible in the CAPO controller logs) include:
+
+- Insufficient quota in the target OpenStack project (for servers, volumes etc.)
+
+- Network routing issues between the CAPI management cluster and the workload cluster
+
+### Essential Addons
+
+Now that the first control plane node has initialised successfully, the kubeconfig for the workload cluster will be written to a Kubernetes `Secret` named `<cluster-name>-kubeconfig` on the CAPI management cluster. At this point, the kubeconfig can be extracted from the secret for use with `kubectl` using the following command:
```
mkjpryor, line 136: This is slightly misleading maybe? The `kubeconfig` secret is actually generated once the API server address is known, which is usually once the LB is created and the FIP attached (if requested).

charts/openstack-cluster/DEBUGGING.md
```diff
@@ -0,0 +1,172 @@
...
+
+```
+kubectl --kubeconfig capi-mgmt get secret <cluster-name>-kubeconfig -o jsonpath='{.data.value}' | base64 -d > workload
```
mkjpryor, line 139: The `go-template` format for `kubectl` supports a `base64decode` filter which I think is nicer than this:

```sh
kubectl --kubeconfig capi-mgmt get secret <cluster-name>-kubeconfig -o go-template='{{.data.value | base64decode}}' > workload
```

Not massively bothered either way though.

mkjpryor (changes requested):
This is a good first crack.

I would maybe just be really explicit and add some examples of accessing the component logs. If you use the `deploy/<>` format, then it can be generic, e.g.:

```sh
kubectl -n capo-system logs deploy/capo-controller-manager`
```

or 

```sh
kubectl -n capi-addon-system logs deploy/cluster-api-addon-provider
```

mkjpryor (approved):
LGTM

PR #421, opened 2024-10-04
added note about using SOURCE_IP_PORT for the lb-method with OVN

gc-nathanh:
The docs currently provide configuration for using the OVN load-balancer provider, but the default lb-method provided by CCM is `ROUND_ROBIN`, which is not compatible with OVN based load balancers.   When attempting to use OVN with `ROUND_ROBIN`, CCM will error with:
`"Provider 'ovn' does not support a requested option: OVN provider does not support ROUND_ROBIN algorithm"`


mkjpryor (approved):
LGTM

PR #464, opened 2024-12-06
Drop requirement for project ID field in cloud credentials

sd109:
Looks like this was fixed upstream at some point since the Gophercloud [type definition](https://github.com/gophercloud/gophercloud/blob/master/openstack/config/clouds/types.go#L86) now allows the project ID to be omitted.

### DIFF
--- a/.github/values/arcus/base.yaml
+++ b/.github/values/arcus/base.yaml
@@ -1,7 +1,5 @@
 clouds:
   arcus:
-    auth:
-      project_id: f0dc9cb312144d0aa44037c9149d2513
     verify: false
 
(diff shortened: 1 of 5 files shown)


sd109:
Upgrade test is failing [here](https://github.com/azimuth-cloud/capi-helm-charts/actions/runs/12200983873/job/34044509992?pr=464#step:11:48) after removing the project_id field from the CI because it starts by trying to install latest published tag which still has hard fail on missing project ID.

mkjpryor (approved):
LGTM. Failed upgrade test is expected.

PR #469, opened 2024-12-17
Allow OCCM to manage LB security groups

sd109:
Workaround for https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2333

PR #471, opened 2024-12-18
Revert "Allow OCCM to manage LB security groups (#469)"

sd109:
It turns out that the OpenStack cloud-controller-manager is not very good at managing security groups, particularly when cluster nodes are deleted and recreated. We've decided that a better solution for now is to revert this and instead pin CAPO to an older working version.


mkjpryor (approved):
LGTM

PR #474, opened 2025-01-13
Enable OCCM-managed LB security groups

sd109:
The previous issues we were seeing with OCCM-managed security groups for OVN LBs were fixed in https://github.com/kubernetes/cloud-provider-openstack/pull/2705 so we can reinstate this fix. Unfortunately OCCM CI is currently broken which means the Helm repos don't have chart version 2.31.2 (which contains the bug fix) available yet so will mark this as DNM for now. I've raised the Helm repo issue in the Kubernetes provider-openstack Slack channel but no timeline yet for fixing their CI.

### DIFF
--- a/charts/cluster-addons/values.yaml
+++ b/charts/cluster-addons/values.yaml
@@ -74,6 +74,8 @@ openstack:
     # By default, ignore volume AZs for Cinder as most clouds have a single globally-attachable Cinder AZ
     BlockStorage:
       ignore-volume-az: true
+    LoadBalancer:
+      manage-security-groups: true
   # Settings for the Cloud Controller Manager (CCM)
   ccm:
     # Indicates if the OpenStack CCM should be enabled


cloudfarm-tech:
Hi,
We just faced with problem of accessibility k8s services types LoadBalancers when with OVN Octavia driver. 
Some researches leaded me to this PR as a permanent fix of it. 
Without **manage-security-groups,** `spec.loadBalancerSourceRanges` is silently discarded under the ovn provider

Components' versions we use in our setup are:

- capi-helm-charts  | 0.21.0
- OCCM helm chart | 2.32.0
- CAPO | v0.14.1
- magnum-capi-helm  | 1.3.1.dev8 


@sd109 @m-bull  is it possible to un-DNM'ing and merging this?

I'm ready to test the rebased chart on our cluster and report back if that helps.

noonedeadpunk:
It would be also nice if somebody could start up CI at least :)

Issue #476, opened 2025-01-20
Values.dnsNameservers no longer respected


dalees:
Commit https://github.com/azimuth-cloud/capi-helm-charts/commit/1628538393b77c4040bdcd0223a3dafab0ff6d3d appears to introduce a bug where the value of `dnsNameservers` is no longer used.

This causes problems when not using `networkFilter` or `subnetFilter`, as the new code will write `managedSubnets.[].Cidr` correctly, but is not able to add any `.dnsNameservers` due to the current context being inside `.Values.internalNetwork`.

This causes creation failures as network DNS values are required. It also causes upgrade failures as the removal of this is seen by OpenStackCluster validation webhook as an invalid change.

The correct context for `dnsNameservers` is `.Values.dnsNameservers`. Can we move these in a backwards compatible way, or do we need to remove the context `with .internalNetwork`?

PR #477, opened 2025-01-20
Update dnsNameservers context to pick up values

dalees:
The `dnsNameservers` value has moved into `managedSubnets` section, but was trying to reference `.internalNetwork.dnsNameservers` which does not exist in chart values.

This change removes the context switch into `internalNetwork` to allow all references to correctly find their related values.

Closes: https://github.com/azimuth-cloud/capi-helm-charts/issues/476

### DIFF
--- a/charts/openstack-cluster/values.yaml
+++ b/charts/openstack-cluster/values.yaml
@@ -73,7 +73,8 @@ kubeNetwork:
 # Settings for the OpenStack networking for the cluster
 clusterNetworking:
   # Custom nameservers to use for the hosts
-  dnsNameservers:
+  # Deprecated; use `clusterNetworking.internalNetwork.dnsNameservers` instead.
+  # dnsNameservers: []
   # Indicates if security groups should be managed by the cluster
   manageSecurityGroups: true
   # Indicates if the managed security groups should allow all in-cluster traffic
@@ -97,6 +98,9 @@ clusterNetworking:
     # The CIDR to use if creating a cluster network
     # This is only used if neither of networkFilter and subnetFilter are given
     nodeCidr: 192.168.3.0/24
+    # List of nameserver IPs to use when creating a cluster network.
+    # This is only used if neither of networkFilter and subnetFilter are given
+    dnsNameservers:
 
 # Settings for registry mirrors
 # When a mirror is set, it will be tried for images but will fall back to the
--- a/charts/openstack-cluster/templates/cluster-openstack.yaml
+++ b/charts/openstack-cluster/templates/cluster-openstack.yaml
@@ -36,7 +36,7 @@ spec:
   {{- else }}
   managedSubnets:
     - cidr: {{ .nodeCidr }}
-      {{- with .dnsNameservers }}
+      {{- with (default $.Values.clusterNetworking.dnsNameservers .dnsNameservers) }}
       dnsNameservers: {{ toYaml . | nindent 8 }}
       {{- end }}
   {{- end }}


dalees:
@mkjpryor If there is a better approach here than removing the context change `with`, please let me know. I'll be happy to update this PR.

It would be nice to move the `.Values` field, but I'm not sure how to do that in Helm, in a backwards compatible way.

mkjpryor:
@dalees 

I was also thinking that the fix here should probably be to move the values field into `internalNetwork`, since it is only respected by the managed subnet.

In order to maintain backwards compatibility, I think we just have to replace

```
{{- with .dnsNameservers }}
```

with

```
{{- with (default $.Values.clusterNetworking.dnsNameservers .dnsNameservers) }}
```

?

dalees:
Ah thanks, @mkjpryor - that was a lot easier than what I was thinking of.

I wasn't sure of the best practice around referencing values outside the current scope.

Updated the PR to use your suggestion and moved the values field.

charts/openstack-cluster/values.yaml
```diff
@@ -73,7 +73,8 @@ kubeNetwork:
 # Settings for the OpenStack networking for the cluster
 clusterNetworking:
   # Custom nameservers to use for the hosts
-  dnsNameservers:
+  # Deprecated; use `clusterNetworking.internalNetwork.dnsNameservers` instead.
+  dnsNameservers: []
```
mkjpryor, line 77: I think just remove this completely

charts/openstack-cluster/values.yaml
```diff
@@ -97,6 +98,10 @@ clusterNetworking:
     # The CIDR to use if creating a cluster network
     # This is only used if neither of networkFilter and subnetFilter are given
     nodeCidr: 192.168.3.0/24
+    # List of nameserver IPs to use when creating a cluster network.
+    # This is only used if neither of networkFilter and subnetFilter are given
+    dnsNameservers: []
```
mkjpryor, line 103: Can we default this to `null` as before, instead of `[]`?

mkjpryor (changes requested):

charts/openstack-cluster/values.yaml
dalees: Perhaps? It would keep the values file cleaner, but I note that the `magnum-capi-helm` driver still uses this - I figured it would be nicer to keep it documented? (Commenting out the actual field may help?)

This appears to be the practice [in other helm charts](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L4232), but happy to change if the code style here differs for a reason. I'm just interested in understanding more.

charts/openstack-cluster/values.yaml
dalees: Annoying style question:

Functionally this is no different (`default` deals with empty lists as if they are `null`) but I felt it provided a hint to the user of the chart as to the type of field this is expected to be.

As before, happy to make the change just interested in why :) It would match the rest of the file better, so that's reason enough perhaps.

charts/openstack-cluster/values.yaml
mkjpryor: Maybe commented out is good.

charts/openstack-cluster/values.yaml
mkjpryor: I was just going for consistency with the previous default, TBH. Not specifying it means "use the network default DNS servers", so I thought null was a better default than `[]` at the time. I'm not that bothered though.

charts/openstack-cluster/values.yaml
dalees: Done

sd109 (approved):
LGTM, thanks for the fix.

PR #481, opened 2025-01-23
Update OpenStackCluster with v1beta1 case and omitempty.

dalees:
These changes match resource changes in CAPO and allow tools such as ArgoCD to fully synchronize the resource instead of showing minor changes. No functionality is changed.

* disableAPIServerFloatingIP is now `omitempty`[1], and is therefore absent when querying the resource when set to 'false'.
* `allowedCidrs` becomes `allowedCIDRs` [2]

[1] https://github.com/kubernetes-sigs/cluster-api-provider-openstack/commit/9223883df61151215286df3442ee1c29a1590c38
[2] https://cluster-api-openstack.sigs.k8s.io/topics/crd-changes/v1alpha7-to-v1beta1#changes-to-ports

### DIFF
--- a/charts/openstack-cluster/templates/cluster-openstack.yaml
+++ b/charts/openstack-cluster/templates/cluster-openstack.yaml
@@ -51,13 +51,15 @@ spec:
     provider: {{ .loadBalancerProvider }}
     {{- end }}
     {{- if .allowedCidrs }}
-    allowedCidrs:
+    allowedCIDRs:
       {{- range .allowedCidrs }}
       - {{ . }}
       {{- end}}
     {{- end }}
   {{- end }}
-  disableAPIServerFloatingIP: {{ not .associateFloatingIP }}
+  {{- if not .associateFloatingIP }}
+  disableAPIServerFloatingIP: true
+  {{- end }}
   {{- with .floatingIP }}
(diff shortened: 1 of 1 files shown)


mkjpryor (approved):
LGTM

mkjpryor:
@dalees 

Looks like this caused the chart upgrade test to fail, which tests upgrading the chart while keeping the same values. This is because the `OpenStackCluster.spec` changes, even though it is effectively a null change (which does annoy me...). I suspect the issue is the `disableAPIServerFloatingIP` change, and I'm not sure how we fix it.

FWIW, the charts as they are cause us no issues with Flux. I'm inclined not to break on chart upgrade for the sake of one tool right now...

mkjpryor (changes requested):
@dalees 

Looks like this caused the chart upgrade test to fail, which tests upgrading the chart while keeping the same values. This is because the `OpenStackCluster.spec` changes, even though it is effectively a null change (which does annoy me...). I suspect the issue is the `disableAPIServerFloatingIP` change, and I'm not sure how we fix it.

FWIW, the charts as they are cause us no issues with Flux. I'm inclined not to break on chart upgrade for the sake of one tool right now...

sd109:
@dalees is this still relevant?

PR #502, opened 2025-03-14
Added CEL lockdown as cluster addon

wtripp180901:
Cluster addon chart now has option to install https://github.com/stackhpc/restricted-pss-vaps-helm/tree/main to apply ValidatingAdmissionPolicy's which define configurable restricted-like pod security standards to the specified namespaces (disabled by default).
By default all capi, kube-system and addon namespaces are excluded from these policies


charts/cluster-addons/values.yaml
```diff
@@ -367,6 +367,38 @@ nodeProblemDetector:
...
+          - node-feature-discovery
+          - gpu-operator
+          - network-operator
+          - node-problem-detector
+          - cilium-secrets
+          - kube-node-lease
+          - kube-public
+          - oidc-system
+          - calico-system
+          - calico-apiserver
```
JohnGarbutt, line 400: Lets add the intel GPU operator in here.

charts/cluster-addons/values.yaml
```diff
@@ -367,6 +367,38 @@ nodeProblemDetector:
     namespace: node-problem-detector
     values: {}
 
+restrictedAdmission:
+  enabled: false
+  chart:
+    repo: https://stackhpc.github.io/restricted-pss-vaps-helm/
+    name: restricted-pss-vaps
+    version: 0.3.0
```
JohnGarbutt, line 375: we need to make sure we auto update this:
https://github.com/azimuth-cloud/capi-helm-charts/blob/dd78906a9939932ecff01ffe254b344a37b3a967/.github/workflows/update-addons.yml#L69

JohnGarbutt (approved):
This looks good now, thank you.

PR #508, opened 2025-03-28
Add Kubernetes 1.32 to upgrade test

sd109:


sd109:
Since GH workflow uses `pull_request_target` this change doesn't actually get tested in CI until it is part of `main` so merging as-is and then will trigger another one of the pending CI runs to validate it works as expected.

PR #509, opened 2025-03-31
Test k8s 1.32 by default in CI

sd109:


sd109:
Can't test properly until merged to main due to `pull_request_target` trigger in relevant GH workflows. Will follow up with separate fix PR after testing if required.

PR #510, opened 2025-04-02
Add node group name validation during templating

sd109:
Example templating errors:

Invalid character:
```
% helm template test charts/openstack-cluster -f values-test.yml --set "nodeGroups[0].name=abc.d"
Error: execution error at (openstack-cluster/templates/node-group/machine-deployment.yaml:4:3): Node group name must be at least three characters long and must contain only lower-case alphanumeric characters and dashes (found name: abc.d)
```

Name too short:
```
% helm template test charts/openstack-cluster -f values-test.yml --set "nodeGroups[0].name=ab"
Error: execution error at (openstack-cluster/templates/node-group/machine-deployment.yaml:4:3): Node group name must be at least three characters long and must contain only lower-case alphanumeric characters and dashes (found name: ab)
```


m-bull (approved):
LGTM

PR #511, opened 2025-04-07
Longhorn addon

assumptionsandg:


charts/cluster-addons/values.yaml
```diff
@@ -357,6 +357,24 @@ nodeProblemDetector:
     namespace: node-problem-detector
     values: {}
 
+# Settings for Longhorn
+longhorn:
+  enabled: true
```
sd109, line 362: Are you sure we want this enabled by default?

charts/cluster-addons/values.yaml
```diff
@@ -357,6 +357,24 @@ nodeProblemDetector:
...
+    repo: https://charts.longhorn.io
+    name: longhorn
+    version: v1.8.0
+  release:
+    namespace: longhorn-system
+    values:
+      persistence:
+        defaultNodeSelector:
+          enable: true
+          selector: "capi.stackhpc.com: dawn"
```
sd109, line 373: Is there a more sensible default we can use here?

charts/cluster-addons/values.yaml
```diff
@@ -359,21 +359,14 @@ nodeProblemDetector:
...
   release:
     namespace: longhorn-system
-    values:
-      persistence:
-        defaultNodeSelector:
-          enable: true
-          selector: "capi.stackhpc.com: dawn"
-      defaultSettings:
-        defaultDataPath: /mnt/nvme
-        defaultDataLocality: strict-local
```
JohnGarbutt: We might want a middle ground where we leave a nice example in the values... Just because I suspect the default settings are not what people want here.

PR #513, opened 2025-04-08
Added Falco addon with default rules for users getting root

wtripp180901:


charts/cluster-addons/values.yaml
```diff
@@ -191,6 +191,90 @@ etcdDefrag:
     namespace: kube-system
     values: {}
 
+falco:
+  enabled: false
+  chart:
+    repo: https://falcosecurity.github.io/charts
```
JohnGarbutt, line 197: We need to add the image sync logic in here as well I think.

PR #519, opened 2025-04-29
Remove references to Argo CD

### LINKED ISSUE #454: What is ControllerReferencesOnly=true? Argocd does not mention it.
The argocd [docs for sync](https://argo-cd.readthedocs.io/en/latest/user-guide/sync-options/) does not mention the `argocd.argoproj.io/sync-options: ControllerReferencesOnly=true` used in those charts? 

What is that option and how is it supposed to work?

sd109:
Closes #454


JohnGarbutt (commented):
We do have some people using the helm charts with Argo, so not sure if they still need this or not.

JohnGarbutt:
> We do have some people using the helm charts with Argo, so not sure if they still need this or not.

Sorry, I have more context now, this is only in that patch, lets remove it.

JohnGarbutt (approved):

PR #521, opened 2025-05-01
Fix whitespace in node group name check

sd109:
Previous version lead to a mangled `helm template` output when two or more node groups were specified.

Old template output included:
```
      nodeDeletionTimeout: 5m0s---
apiVersion: cluster.x-k8s.io/v1beta1
kind: MachineDeployment
```

fixed version templates correctly as:
```
      nodeDeletionTimeout: 5m0s
---
# Source: openstack-cluster/templates/node-group/machine-deployment.yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: MachineDeployment
```


JohnGarbutt:
@sd109 can we add that test that validates the output of helm template? I am guessing that would pick it up?

sd109:
Yeah I'm hoping we can use something like [helm unittest snapshots](https://github.com/helm-unittest/helm-unittest?tab=readme-ov-file#snapshot-testing) to catch these kinds of bugs. I'll look into it as part of the Helm CI improvements over the next few weeks.

wtripp180901 (approved):
tested in dev environment and seems to have fixed the node group issue

JohnGarbutt (approved):
Sounds good Scott, lets get this merged.

PR #543, opened 2025-05-08
Fix bug in DNS nameserver values

Alex-Welsh:
The DNS nameservers option was in the wrong place in the values.yml file, where it should have been in the internal network block: https://github.com/azimuth-cloud/capi-helm-charts/blob/65c92d45fd76db299bb318e46377cd03107dff53/charts/openstack-cluster/templates/cluster-openstack.yaml#L27-L40

### DIFF
--- a/charts/openstack-cluster/values.yaml
+++ b/charts/openstack-cluster/values.yaml
@@ -72,8 +72,6 @@ kubeNetwork:
 
 # Settings for the OpenStack networking for the cluster
 clusterNetworking:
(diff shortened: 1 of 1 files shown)


sd109:
I think CI is failing due to unrelated cloud-level networking issues (as discussed elsewhere). Will retry CI run once cloud issues are resolved.

sd109:
I've just realised that this was already raised in #476 and WIP fix is in #477 including backwards compatibility considerations so we should close this PR and get that fix in instead.

PR #546, opened 2025-05-13
Add chart manifest validation tests

sd109:
Adds two extra steps to to validate Helm chart content. 

First uses `helm template | kubeconform` to validate the templated manifests against the published OpenAPI schema for the k8s resource types. 

Second uses the `helm unittest` plugin to generate manifest snapshots and store them under version control to allow easier review of template changes and avoid changing templates in unintentional ways which break backwards compatibility.

Workflows changes won't be tested until merged to main but I've pushed the changes to a fork to test and demo the workflows.

Example of successful test run:
https://github.com/sd109/capi-helm-charts/actions/runs/14998199202/job/42137726412

Example of how `kubeconform` test would have caught bug fixed in #521:
https://github.com/sd109/capi-helm-charts/actions/runs/14995851028/job/42129764254

Example of how updating manifest snapshots can make it easier to review template changes in PRs:
https://github.com/azimuth-cloud/capi-helm-charts/pull/537

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -7,7 +7,6 @@ on:
         description: The Git ref under test.
         required: true
 
-
 jobs:
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -6,6 +6,14 @@ on:
         type: string
         description: The Git ref under test.
         required: true
+  # TODO: Remove after testing
+  workflow_dispatch:
+    inputs:
+      ref:
+        type: string
+        description: The Git ref under test.
+        required: true
+
 
 jobs:
   lint:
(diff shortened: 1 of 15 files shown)


.github/workflows/lint.yaml
```diff
@@ -36,3 +31,17 @@ jobs:
             --target-branch ${{ github.event.repository.default_branch }} \
             --all \
             --validate-maintainers=false
+
+      - name: Run template validation
```
m-bull, line 35: Is it worth making this a reusable workflow that we could call from places that template capi-helm-charts values?

m-bull: On second thoughts, we could just add the same test to places that template capi-helm-charts values.yaml, like the operator and azimuth-ops

m-bull (commented):
I think this seems good - it certainly seems sensible to try and catch the kind of errors in the desciription.

JohnGarbutt (approved):
This looks awesome, thank you.

PR #551, opened 2025-05-21
Added support for azimuth-authorization-webhook

wtripp180901:
Installs config files via KubeADM that allows clusters to use https://github.com/azimuth-cloud/azimuth-authorization-webhook as an authorisation webhook


wtripp180901:
+1

charts/openstack-cluster/values.yaml
```diff
@@ -190,6 +190,8 @@ apiServer:
   fixedIP:
   # The port to use for the API server
   port: 6443
+  # Log level
```
JohnGarbutt, line 193: Might be worth giving some example values here?

charts/openstack-cluster/values.yaml
```diff
@@ -214,8 +216,28 @@ osDistro: ubuntu
...
+  webhookVersion: v1
+  timeout: 10s
+  failurePolicy: Deny
+  additionalAuthorizers:
+    - type: Node
+      name: node
+    - type: RBAC
+      name: rbac
+  tls:
+    enabled: false
```
JohnGarbutt, line 233: Maybe, lets make this true, and just make it clear you really need to specify the cert for anything to work? Not sure.

JohnGarbutt (approved):
Great stuff, thank you.

PR #553, opened 2025-05-23
Drop K8s 1.29 now its out of support

JohnGarbutt:
This is need so the tests pass on the latest
azimuth images release.


m-bull:
Unless you want to include it here, we should make a note to include 1.33 CI in a follow-up

sd109 (changes requested):
I think we need to add the 1.33 equivalents in all of the places where you've removed the 1.29 references.

Here are the two PRs that I made to do the same for 1.32. I did it in two PRs because I missed things the first time but we should do it all in one PR here.
- https://github.com/azimuth-cloud/capi-helm-charts/pull/508
- https://github.com/azimuth-cloud/capi-helm-charts/pull/509

sd109:
Ah sorry, I didn't refresh the page before reviewing so missed @m-bull's comment, but I think we should do the 1.33 as part of this PR so that we have it all in one please for future reference.

PR #559, opened 2025-06-10
Added etcd encryption support

wtripp180901:


wtripp180901:
NB: we might want to do some refactoring before this gets merged. The apiserver gets patched here and also in helpers.tpl (which at the moment is designed to be specifically for patching in webhooks), which leaves a bit of a footgun in that it's pretty easy to make patch files with duplicate names and not realise if the sets of components requiring the patches aren't deployed together

charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
```diff
@@ -69,6 +69,61 @@ mounts:
...
+            {{- .Values.etcd.encryption.resources | toYaml | nindent 10 }}
+          providers:
+            - {{ .Values.etcd.encryption.provider }}:
+                keys:
+                  - name: key1
+                    {{- $secret := (lookup "v1" "Secret" .Release.Namespace (print .Release.Name "-etcd-key")) -}}
+                    {{- if $secret }}
+                    secret: {{ $secret.data.key }}
+                    {{- else }}
+                    secret: "Fallback value for CI only"
```
JohnGarbutt, line 119: This seems a bit dangerous, can we fix up CI so it has a secret created that can be used?

JohnGarbutt: Lets add the context in a comment if we can, that sounds like a good call.

wtripp180901:
off by default now, moved unfinished CI to https://github.com/azimuth-cloud/capi-helm-charts/pull/596

PR #564, opened 2025-06-30
Update manifest snapshots

sd109:
Minor fixes which weren't caught previously since Helm lint CI uses pull request target trigger so wasn't properly tested until #546 was merged.


sd109:
Linting job now passes on main: https://github.com/azimuth-cloud/capi-helm-charts/actions/runs/15977692689/job/45064326008

PR #600, opened 2025-07-16
Keystone Auth: Remove authorization-* args from apiserver bootstrap

dalees:
In kubernetes 1.29 kubeadm is not able to authorize the admin user because it is trying to find a webhook that doesn't yet exist.

Solution: don't add webhook at init time but instead wait till after kubeadm has completed and then use kustomize to patch the kube-apiserver with the authorization-mode argument to add "Webhook". It is possible to add this option to the kube-apiserver arguments separately so it can just be appended to the list of args along with the authorization/authentication config paths.

Related issue: https://github.com/kubernetes/cloud-provider-openstack/issues/2575

### DIFF
--- a/charts/openstack-cluster/templates/_helpers.tpl
+++ b/charts/openstack-cluster/templates/_helpers.tpl
@@ -392,9 +392,16 @@ ignition:
 Create folders necessary for webhook integration.
 */}}
 {{- define "openstack-cluster.webhookPatches" }}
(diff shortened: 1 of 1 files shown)


dalees:
Related discussion in Kubernetes Slack, in #openstack-magnum: https://kubernetes.slack.com/archives/C0639B2CX5E/p1750892690223919

PR #609, opened 2025-08-20
Replace unmaintained loki-stack addon with separate loki and promtail addons

### LINKED ISSUE #606: Loki-Stack is Deprecated
The [Loki-Stack chart is deprecated](https://community.grafana.com/t/difference-between-helm-charts-loki-and-loki-stack/87380/2), it is using outdated versions of loki/promtail/fluentbit (if enabled).

[Promtail is also deprecated](https://grafana.com/docs/loki/latest/send-data/promtail/#promtail-agent) in favor of Alloy.

It should be switched out for the updated version of the normal Loki-Chart, and either Alloy/fluent-operator for log collection.

This, for example, means we can't use Drilldown -> Logs in Grafana as it expects a more up-to-date version of Loki which allows for log volume queries.

As for fluent-operator, we've already switched promtail out for it [here](https://github.com/stfc/cloud-azimuth-config/blob/main/environments/stfc-base/inventory/group_vars/all/user_cluster_addons.yml) and [here](https://github.com/stfc/cloud-azimuth-config/blob/main/environments/stfc-base/inventory/group_vars/all/user_cluster_fluent_operator_values.yml) if it helps give you reusable segments of code.

We haven't attempted to fully switch out loki-stack; we'd need to do it with custom addons and there's not much control possible there over i.e. PVC hooks which is used in the [loki-stack cluster addon](https://github.com/azimuth-cloud/capi-helm-charts/blob/main/charts/cluster-addons/templates/monitoring/loki-stack.yaml).

JasleenKaurSethi:
Closes #606

### DIFF
--- a/charts/cluster-addons/values.yaml
+++ b/charts/cluster-addons/values.yaml
@@ -319,7 +319,6 @@ monitoring:
     release:
       namespace: monitoring-system
       values: {}
-      
   # Note(sd109): Loki v2 is the successor to the lokiStack addon since the Helm
   # chart for the latter is no longer maintained upstream.
 


sd109:
Closes #606

sd109:
> Promtail is also deprecated, is the plan to initially switch out loki-stack for loki then switch promtail -> alloy later?

Yes exactly, we're planning to get to latest Loki v2.X first and then follow up with a separate change to upgrade to Loki v3.X which is where the migration from Promtail to Alloy will happen. We have a few long-term deployments which make use of the existing logging stack so we're aiming to make the switch as smooth as possible and avoid losing any historical data if possible, hence the more gradual approach.

charts/cluster-addons/templates/monitoring/loki-stack.yaml
```diff
@@ -1,86 +1,95 @@
...
-          readOnly: true
-    grafana:
-      sidecar:
-        datasources:
-          enabled: false
+        serviceMonitor:
+          enabled: true
+        # Get promtail to scrape systemd services
+        config:
+          snippets:
```
sd109, line 33: It looks like we've got two layers of `config` key here now - is this correct? The values now look like
```yaml
promtail:
  config:
    config:
      snippets:
        ...
```

skopeo-manifests/etcd-defrag.yaml
```diff
@@ -1,7 +1,4 @@
 docker.io:
   images:
-    bitnami/kubectl:
-    - "1.30"
-    - "1.31"
-    - "1.32"
-    - "1.33"
+    alpine/kubectl:
+    - "1.34.1"
```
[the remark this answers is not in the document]
JasleenKaurSethi: I don't think so, when I try to pull bitnami using 1.30 version for instance, it gives an error saying "Error response from daemon: manifest for alpine/kubectl:1.30 not found: manifest unknown: manifest unknown"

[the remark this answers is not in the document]
sd109: If you merge or rebase to get latest changes from main branch you'll get [this fix](https://github.com/azimuth-cloud/capi-helm-charts/pull/627) which should resolve the kubectl issue in CI.

charts/cluster-addons/templates/monitoring/loki-stack.yaml
```diff
@@ -1,86 +1,101 @@
...
+  name: {{ include "cluster-addons.componentName" (list . "promtail") }}-config
+  labels:
+    {{- include "cluster-addons.componentLabels" (list . "promtail") | nindent 4 }}
+    addons.stackhpc.com/watch: ""
+stringData:
+  defaults: |
+    config:
       serviceMonitor:
         enabled: true
-      # Get promtail to scrape systemd services
```
sd109: We seem to have lost this explanatory comment in the new config below, can we re-add it please?

charts/cluster-addons/templates/monitoring/loki-stack.yaml
```diff
@@ -1,86 +1,101 @@
...
-          readOnly: true
-    grafana:
-      sidecar:
-        datasources:
-          enabled: false
+      clients:
+        {{- if not (and .Values.monitoring.promtail.release.values.config (and .Values.monitoring.promtail.release.values.config.clients (ne (index .Values.monitoring.promtail.release.values.config.clients 0).url ""))) }}
+        - url: http://loki-stack.{{ .Values.monitoring.loki.release.namespace }}.svc:3100/loki/api/v1/push
+          tenant_id: 1
+        {{- end }}
```
sd109: Why do we need this conditional? I think we can probably just drop it and have
```suggestion
        - url: http://loki-stack.{{ .Values.monitoring.loki.release.namespace }}.svc:3100/loki/api/v1/push
          tenant_id: 1
```
If a user of these charts wants to add their own list of clients they are free to overwrite it completely using `.Values.monitoring.loki.release.values`. I think trying to second-guess the user and add client URLs conditionally is more likely to just cause confusion.

charts/cluster-addons/templates/monitoring/loki-stack.yaml
```diff
@@ -102,31 +117,44 @@ spec:
...
+  targetNamespace: {{ .Values.monitoring.promtail.release.namespace }}
+  releaseName: promtail
+  valuesSources:
+    - secret:
+        name: {{ include "cluster-addons.componentName" (list . "promtail") }}-config
+        key: defaults
+    - secret:
+        name: {{ include "cluster-addons.componentName" (list . "promtail") }}-config
+        key: overrides
+{{- end }}
```
sd109: If this `{{- end }}` is for the `if` clause in line 1 of the file then I think should be add the bottom of the file? At the moment I think the loki-dashboards Manifest is included even if Loki is disabled which is not what we want.

charts/cluster-addons/values.yaml
```diff
@@ -309,6 +309,79 @@ monitoring:
...
+            replication_factor: 1
+            # Specify path for new Loki to use same path in PVC
+            # as old loki-stack for reading previously logged data
+            # required in case historical logs are needed
+            # path_prefix: /var/loki/loki
+          storage:
+            type: 'filesystem'
+            filesystem:
+              chunks_directory: /var/loki/loki/chunks
+              rules_directory: /var/loki/loki/rules
```
sd109: Is this also part of the config that's only required for keeping historical logs during upgrade?

.github/workflows/update-addons.yml
```diff
@@ -60,9 +60,15 @@ jobs:
           - key: kube-prometheus-stack
             path: monitoring.kubePrometheusStack.chart
 
+          - key: loki
+            path: monitoring.loki.chart
```
sd109:
```suggestion
            path: monitoring.loki_v2.chart
```

charts/cluster-addons/templates/monitoring/loki.yaml
```diff
@@ -0,0 +1,123 @@
...
+  labels:
+    {{- include "cluster-addons.componentLabels" (list . "loki") | nindent 4 }}
+    addons.stackhpc.com/watch: ""
+stringData:
+  defaults: |
+    serviceMonitor:
+      enabled: true
+    monitoring:
+      dashboards:
+        enabled: false
```
sd109: According to https://artifacthub.io/packages/helm/grafana/loki/5.48.0?modal=values&path=monitoring.serviceMonitor the `serviceMonitor` section should be under the `monitoring` key:
```suggestion
  defaults: |
    monitoring:
      serviceMonitor:
        enabled: true
      dashboards:
        enabled: false
```

charts/cluster-addons/templates/monitoring/loki.yaml
```diff
@@ -0,0 +1,123 @@
...
+                  storage: {{ .Values.monitoring.loki_v2.release.values.singleBinary.persistence.size }}
+    {{- else }}
+    postUpgrade:
+      - apiVersion: v1
+        kind: PersistentVolumeClaim
+        selector:
+          app: loki
+          release: loki-stack
+        action: delete
+    {{- end }}
```
sd109: Can you test that these hooks still work with the new chart please? I think the process should be:
1. Deploy cluster with loki_v2 enabled
2. Increase the persistence.size config value then run a `helm upgrade`
3. Check that the PVC and corresponding Cinder volume get correctly resized
4. Set persistence.enabled = false and run another Helm upgrade
5. Check that the PVC and corresponding Cinder volume get deleted
6. Delete the test cluster

charts/cluster-addons/values.yaml
```diff
@@ -309,6 +309,81 @@ monitoring:
...
+            # path_prefix: /var/loki/loki
+          storage:
+            type: 'filesystem'
+            # Required configuration to store data in /var/loki
+            # if historical data is needed
+            # filesystem:
+            #   chunks_directory: /var/loki/loki/chunks
+            #   rules_directory: /var/loki/loki/rules
+          # Schema configuration needed in case historical logs are needed
+          # schemaConfig:
```
sd109:
```suggestion
          # If historical logs are required, additional schema configuration 
          # is needed to allow loki to ingest old loki-stack logging data.
          # For example:
          # schemaConfig:
```

PR #698, opened 2026-04-10
Restore write permissions to GITHUB_TOKEN

m-bull:
Since setting all GITHUB_TOKEN to read-only at the Org-level, some jobs need their implicit write permissions restored.


m-bull:
I don't expect this to pass checks because of pull_request_target

wtripp180901 (approved):

PR #700, opened 2026-04-13
Extend GITHUB_TOKEN permissions to jobs calling workflows 

m-bull:


m-bull:
Again, because of pull_request_target, CI jobs here will never run

wtripp180901 (approved):

PR #712, opened 2026-04-30
Add switch for default Openstack Blackbox target

wtripp180901:


charts/cluster-addons/templates/monitoring/blackbox-exporter.yaml
```diff
@@ -11,7 +11,7 @@ stringData:
   defaults: |
     serviceMonitor:
       enabled: true
-      {% if cloud_identity and "clouds.yaml" in cloud_identity.data %}
+      {% if .Values.monitoring.blackboxExporter.openstackAuthUrlTarget.enabled and cloud_identity and "clouds.yaml" in cloud_identity.data %}
```
m-bull: I don't think you can use Helm values here - this is templated by the addon provider using Jinja (https://github.com/azimuth-cloud/cluster-api-addon-provider/#variables).

Just use this as a switch a couple of lines up to remove the whole of `defaults:` using Helm templating like:

```
{{- if .Values.monitoring.blackboxExporter.openstackAuthUrlTarget.enabled }}
defaults: |
 blah
{{- end }}
```

m-bull (changes requested):

m-bull (approved):
LGTM

PR #741, opened 2026-05-18
Add nodeTaints support to openstack-cluster

sorrison:
## Summary
  - New `nodeTaints` per-worker-node-group key, rendered into the worker
    `KubeadmConfigTemplate` as a native
    `joinConfiguration.nodeRegistration.taints` list
  - New helper `openstack-cluster.nodeRegistration.nodeTaints` and a parallel
    `nodegroup.kct.spec.nodeTaints` define that merges into the existing
    per-group spec alongside the `nodeLabels` define
  - `values.yaml` documents the new key; `tests/values_full.yaml` exercises it
    and snapshots are regenerated
    
  Pairs with a forthcoming change in OpenStack Magnum (microversion 1.13)
  that exposes `node_labels` / `node_taints` on the NodeGroup REST resource,
  and a `magnum-capi-helm` change that forwards those fields as `nodeLabels`
  / `nodeTaints` in the Helm values.
  
  ## Test plan
  - [ ] `helm dependency update charts/openstack-cluster`
  - [ ] `docker run -i --rm -v $(pwd):/apps helmunittest/helm-unittest charts/openstack-cluster`
        (no `-u` — confirm snapshots are clean as-is)
  - [ ] Deploy a cluster with `nodeTaints` set on a node group; verify the
        rendered MachineDeployment's `KubeadmConfigTemplate` contains
        `spec.template.spec.joinConfiguration.nodeRegistration.taints`
  - [ ] Update `nodeTaints` on an existing release; verify a new
        `KubeadmConfigTemplate` name is generated (checksum suffix changes)
        and Cluster API rolls the MachineDeployment


mgrzybek (changes requested):
`values.schema.json` is not regenerated. Why?
The news attribute must be documented.

sorrison:
I'll rebase to take into account recent changes,  this patch was uploaded a while ago before schema existed

mgrzybek:
Hey @sorrison,

I started the pipeline and everything is green. I am waiting for you to resolve conflicts.

charts/openstack-cluster/values.yaml
```diff
@@ -973,6 +977,28 @@ nodeGroupDefaults:
...
+  #   properties:
+  #     key:
+  #       type: string
+  #     value:
+  #       type: string
+  #     effect:
+  #       enum: [NoSchedule, PreferNoSchedule, NoExecute]
+  #   required: [key, effect]
+  # @schema
+  nodeTaints: []
```
mgrzybek, line 449: Minor/informational, not blocking: the values.yaml comment-to-schema-description generator attaches a key's trailing "example" comment (the commented-out sample block right after a key with no inline value) to the description of the *next* key rather than its own. This is a pre-existing quirk (e.g. `nodeLabels`'s example previously bled into `additionalBlockDevices`'s generated description), but inserting `nodeTaints` here between them makes it cascade further: `nodeTaints`'s own description picks up `nodeLabels`'s leftover `my.company.org/label: value` example, and `additionalBlockDevices` right below picks up `nodeTaints`'s `- key: my.company.org/reserved...` example instead.

Since `values.schema.json` was fully reverted from `main` in #806 (this PR predates that revert and needs a rebase — see PR conversation), this is currently moot in practice. Flagging so it's not forgotten if the schema-generation feature is ever reinstated.

sorrison: OK should be all sorted, cheers

mgrzybek (approved):

PR #745, opened 2026-06-02
A node group can contain zero machines

mgrzybek:
The scale to/from 0 is now supported by CAPI. Some work is in progress on Magnum’s side (https://review.opendev.org/c/openstack/magnum-capi-helm/+/989645).

We need to allow this value in Helm and make it available in the next release. Then, `magnum-capi-helm` Wille be able to verify that the correct release in deployed and allow the node group to be set to 0.


wtripp180901 (approved):
LGTM

PR #746, opened 2026-06-03
Update images for node20.

amanning9:


.github/workflows/ensure-capi-images.yaml
```diff
@@ -88,7 +88,7 @@ jobs:
         if: ${{ !matrix.skip }}
 
       - name: Write matrix outputs
-        uses: cloudposse/github-action-matrix-outputs-write@ed06cf3a6bf23b8dce36d1cf0d63123885bb8375 # v1
+        uses: cloudposse/github-action-matrix-outputs-write@ed06cf3a6bf23b8dce36d1cf0d63123885bb8375 # v1.0.0
```
m-bull, line 96: I don't think this is node24

m-bull (changes requested):

.github/workflows/ensure-capi-images.yaml
amanning9: I've been working on this same problem in azimuth-images, so I've set this one to allow node20 for now until that's finished.

m-bull (approved):
LGTM

PR #747, opened 2026-06-12
Configure worker and control-plane security groups

m-bull:
Since CAPO v0.11.0 security groups prevent traffic originating from outside of the cluster network from reaching NodePorts [1]. This caused a regression in the expected behaviour on clouds that were using the OVN Octavia loadbalancer provider [2].

To work around this regression, since CAPO v0.12.0 it is possible to set workerNodeSecurityGroupRules and controlPlaneNodesSecurityGroupRules, containing a list of security group rules that should be applied to worker and control-plane nodes respectively [3].

Set workerNodeSecurityGroupRules when apiServer.loadBalancerProvider=ovn, and expose worker and control-plane security groups in values.yaml.

[1] https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2128
[2] https://github.com/kubernetes-sigs/cluster-api-provider-openstack/issues/2333
[3] https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/2353

### DIFF
--- a/charts/openstack-cluster/templates/_helpers.tpl
+++ b/charts/openstack-cluster/templates/_helpers.tpl
@@ -576,3 +576,39 @@ Produces integration for azimuth_authorization_webhook on apiserver
             name: {{ .name }}
           {{- end }}
 {{- end }}
+
+{{/*
+Return a list of security groups to apply to worker nodes when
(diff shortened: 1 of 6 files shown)


irt-shpc (approved):
LGTM

PR #749, opened 2026-06-19
Remove prekubeadmcommand for hosts.toml paths

m-bull:
Since [1] image-builder has handled adding this configuration to underlying images at build-time, so we don't need to mess with it at runtime.

Also add ORC setup to hacking-setup, which is required post the recent CAPO upgrade.

[1] https://github.com/kubernetes-sigs/image-builder/commit/12c608c3b0aa3c93cb0740ac7f9d30d1d0b6e85d


irt-shpc (approved):

mgrzybek:
Argh, I made a mistake when resolving conflicts. Let’s do it again.

PR #754, opened 2026-06-19
Fix broken invocation of helm-unittest

m-bull:
Fixes issue where helm-unittest container was unable to create `/.cache`

> level=WARN msg="failed to load plugin (ignoring)" plugin_yaml=/usr/local/share/helm/plugins/unittest/plugin.yaml error="failed to create plugin manager: failed to create wazero compilation cache: create directory /.cache/helm/wazero-build: mkdir /.cache: permission denied"
Error: unknown command "unittest" for "helm"


sjpb (approved):

m-bull:
Tested working, @sjpb's review doesn't count towards requirements but this needs to not be broken so bypassing to merge anyway.

PR #762, opened 2026-06-30
.github>upgrade-and-test/action.yml: Test etcd-defrag after every upg…

irt-shpc:
…rade


mgrzybek (approved):

.github/actions/upgrade-and-test/action.yml
```diff
@@ -216,3 +216,16 @@ runs:
...
+
+    - name: Run etcd-defrag job
+      shell: bash
+      run: |-
+        kubectl create job -n kube-system \
+          --from=cronjob/etcd-defrag etcd-defrag &&
+        kubectl wait job/etcd-defrag \
+          -n kube-system \
+          --for=condition=complete \
+          --timeout 3m
```
m-bull: I think the last thing that this needs is a `kubectl delete` at the end, or a unique suffix for the job name. 

Because we do actions on the same cluster (upgrades etc), when we go to create the job again, one already exists with the same name.

m-bull (changes requested):

.github/actions/upgrade-and-test/action.yml
```diff
@@ -216,3 +216,23 @@ runs:
...
+      env:
+        KUBECONFIG: ./kubeconfig
+
+    - name: Delete etcd-defrag job
+      shell: bash
+      run: |-
+        kubectl delete jon -n kube-system \
```
m-bull: Who's Jon!?

😂

m-bull (changes requested):

m-bull (approved):

PR #763, opened 2026-07-01
Expose KCP certificatesExpiryDays in the openstack-cluster chart

jihyun-huh:
## Summary

Expose the KubeadmControlPlane `rolloutBefore.certificatesExpiryDays` option through the `openstack-cluster` Helm chart.

This allows deployments to opt in to Cluster API's automatic control plane rollout before kubeadm-managed certificates expire.

- related CAPI handbook: https://cluster-api.sigs.k8s.io/tasks/certs/auto-rotate-certificates-in-kcp

## Motivation

By default, kubeadm generated control plane certificates expire after one year. If a workload cluster is not upgraded before then, certificate expiry can lead to cluster disruption and manual recovery.

Cluster API has supported this rollout mechanism for some time, but it has not previously been exposed through the Azimuth Helm charts. This change makes the existing upstream capability available without changing the default behaviour.

This change exposes that field so deployments can enable automatic certificate-driven rollouts when appropriate.

## Notes

- The value is optional and unset by default.
- Existing behaviour is unchanged unless the value is explicitly configured.
- This change only exposes existing Cluster API functionality; it does not change the rollout logic itself.
- This wiki is describing why this PR is needed: https://wiki.stackhpc.com/doc/azimuth-automatic-certificate-rotation-in-cluster-api-iEH8Cy8HHH

### DIFF
--- a/charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
+++ b/charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
@@ -179,6 +179,10 @@ spec:
   replicas: {{ .Values.controlPlane.machineCount }}
   remediationStrategy: {{ toYaml .Values.controlPlane.remediationStrategy | nindent 4 }}
   rolloutStrategy: {{ toYaml .Values.controlPlane.rolloutStrategy | nindent 4 }}
+   {{- if gt (.Values.controlPlane.certificatesExpiryDays | int) 0 }}
+  rolloutBefore:
+    certificatesExpiryDays: {{ .Values.controlPlane.certificatesExpiryDays | int }}
+  {{- end }}
   machineTemplate:
     metadata:
       labels: {{ include "openstack-cluster.componentSelectorLabels" (list . "control-plane") | nindent 8 }}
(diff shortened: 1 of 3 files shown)


charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
```diff
@@ -179,6 +179,10 @@ spec:
   replicas: {{ .Values.controlPlane.machineCount }}
   remediationStrategy: {{ toYaml .Values.controlPlane.remediationStrategy | nindent 4 }}
   rolloutStrategy: {{ toYaml .Values.controlPlane.rolloutStrategy | nindent 4 }}
+  {{- if .Values.controlPlane.certificatesExpiryDays }}
```
irt-shpc: This could add some safety with
` {{- if (.Values.controlPlane.certificatesExpiryDays | int) > 0}}`

That might not quite be the right syntax but check the variable is an integer

irt-shpc (changes requested):

charts/openstack-cluster/templates/control-plane/kubeadm-control-plane.yaml
jihyun-huh: thank you for your review! 
and i updated my code with using variable type check

irt-shpc (approved):
LGTM assuming tests pass

PR #770, opened 2026-07-06
feat(flatcar): add sysext download and provider-id to Ignition config

mgrzybek:
Extend the existing `flatcarKubeadmConfigSpec` helper to wire up the full Kubernetes bootstrap on Flatcar Linux via `systemd-sysext`.

- Add `flatcar.sysext{Registry,KubernetesTag,KubernetesDigest, ContainerdTag,ContainerdDigest}` values for addressing OCI artifacts in Harbor
- Inject storage.files (sysext blob download) and storage.links (symlinks into `/etc/extensions/`) into the Ignition `additionalConfig`; Ignition fetches them synchronously before any service starts
- Set `cloud-provider=external` and `provider-id=openstack:///<UUID>` kubelet args via `nodeRegistration`, substituted at runtime by envsubst from the coreos-metadata `EnvironmentFile`
- Export `COREOS_OPENSTACK_INSTANCE_UUID` alongside the existing hostname export so envsubst can resolve both variables in `/etc/kubeadm.yml`

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -486,7 +486,7 @@ jobs:
 
       - name: Write Flatcar override values
         run: |-
-          cat > /tmp/flatcar-overrides.yaml <<EOF
+          cat > /tmp/flatcar-overrides.yaml <<'EOF'
           osDistro: flatcar
           flatcar:
             sysextRegistry: "${{ vars.FLATCAR_SYSEXT_REGISTRY }}"
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -456,15 +456,15 @@ jobs:
           process-id: ${{ github.repository }}/${{ github.run_id }}/${{ github.job }}
         if: ${{ vars.TARGET_CLOUD == 'arcus' && always() }}
 
-  # This job tests a basic Flatcar Linux deployment (Ignition + sysexts)
+  # This job tests a Flatcar Linux deployment using systemd-sysext (Ignition + sysexts)
   # Requires these repository variables to be configured:
-  #   FLATCAR_IMAGE_ID          — Glance UUID of the flatcar-stable base image
-  #   FLATCAR_SYSEXT_REGISTRY   — Harbor hostname serving flatcar sysexts (e.g. harbor.example.com)
-  #   FLATCAR_KUBE_VERSION      — Kubernetes version to test (e.g. 1.35.6)
-  #   FLATCAR_KUBE_DIGEST       — Blob digest for the kubernetes sysext (sha256:...)
-  #   FLATCAR_CONTAINERD_VERSION — containerd sysext version (e.g. 1.7.27)
-  #   FLATCAR_CONTAINERD_DIGEST  — Blob digest for the containerd sysext (sha256:...)
-  flatcar:
+  #   FLATCAR_IMAGE_ID                    — Glance UUID of the flatcar-stable base image
+  #   FLATCAR_KUBE_VERSION                — Kubernetes version to test (e.g. 1.35.6)
+  #   FLATCAR_SYSEXT_KUBERNETES_URL       — S3 public URL for the Kubernetes sysext .raw file
+  #   FLATCAR_SYSEXT_KUBERNETES_CHECKSUM  — SHA256 checksum of the Kubernetes sysext (sha256-...)
+  #   FLATCAR_SYSEXT_CONTAINERD_URL       — S3 public URL for the containerd sysext .raw file
+  #   FLATCAR_SYSEXT_CONTAINERD_CHECKSUM  — SHA256 checksum of the containerd sysext (sha256-...)
+  flatcar-sysext:
     runs-on: ubuntu-latest
     if: ${{ inputs.tests-full && vars.FLATCAR_IMAGE_ID != '' }}
     steps:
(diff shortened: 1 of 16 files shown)


irt-shpc (approved):

irt-shpc:
Needs rebasing

mgrzybek:
We should wait for https://github.com/azimuth-cloud/azimuth-images/pull/482 to be merged and the Flatcar images to be availabe.

charts/openstack-cluster/values.yaml
```diff
@@ -473,6 +473,17 @@ apiServer:
 # Set osDistro used. ubuntu, flatcar, etc.
 osDistro: ubuntu
 
+# Flatcar Linux sysext configuration (only used when osDistro: flatcar)
+flatcar:
```
m-bull, line 239: As there are two options for flatcar in CAPO land, `flatcar` and `flatcar-sysext`, I think we should gate these under `flatcar-sysext`, so that we don't break people using `flatcar` as it is.

That would allow us to only `require` these (https://github.com/azimuth-cloud/capi-helm-charts/pull/770/changes#diff-69445aa57c2a517f040e86a387853612efaacca40edb3ba2988de35a859dcf38R394) urls and checksums when `osDistro` is `flatcar-sysext`, which is backward compatible and won't break people using `osDistro: flatcar`.

m-bull (changes requested):

.github/workflows/test.yaml
```diff
@@ -455,3 +455,90 @@ jobs:
...
+          bucket: ${{ vars.CI_S3_LOCK_BUCKET }}
+          action: acquire
+          process-id: ${{ github.repository }}/${{ github.run_id }}/${{ github.job }}
+        timeout-minutes: 180
+        if: ${{ vars.TARGET_CLOUD == 'arcus' }}
+
+      - name: Test Flatcar sysext deployment
+        uses: ./.github/actions/upgrade-and-test
+        with:
+          name: ci-${{ github.run_id }}-${{ github.job }}
```
m-bull, line 516: It would be nice to at `ci-flatcar-` or something just for tracking

.github/workflows/test.yaml
```diff
@@ -455,3 +455,90 @@ jobs:
           action: release
           process-id: ${{ github.repository }}/${{ github.run_id }}/${{ github.job }}
         if: ${{ vars.TARGET_CLOUD == 'arcus' && always() }}
+
+  # This job tests a Flatcar Linux deployment using systemd-sysext (Ignition + sysexts)
+  # Requires these repository variables to be configured:
```
m-bull, line 460: None of these exist, so we should populate

charts/openstack-cluster/README.md
```diff
@@ -627,12 +627,44 @@ See [DEBUGGING.md](./DEBUGGING.md).
 
 ### Flatcar support
 
-To deploy clusters which use Ignition such as Flatcar, you will need to override the following
-setting in your local `values.yaml`:
+To deploy clusters which use Ignition such as Flatcar, you will need to override the `osDistro`
```
m-bull, line 630:
```suggestion
To deploy clusters based on Flatcar Linux, you will need to override the `osDistro`
```

I think the ignition bit here is too much information - the distro is the important bit, how it is bootstrapped less so

charts/openstack-cluster/README.md
```diff
@@ -627,12 +627,44 @@ See [DEBUGGING.md](./DEBUGGING.md).
...
+setting in your local `values.yaml`. Both variants use
```
m-bull, line 631: Both variants? Define/introduce them upfront (faltcar, flatcar-sysext).

charts/openstack-cluster/README.md
```diff
@@ -627,12 +627,44 @@ See [DEBUGGING.md](./DEBUGGING.md).
...
+```
+Use this with a generic `flatcar-stable` image: Ignition fetches the sysext images
+during first-boot provisioning and writes them under `/opt/extensions/`.
+systemd-sysext then merges them on every boot exactly as it would a baked-in image.
+
+You may add arbitary sysexts to the flatcar.sysexts dict.
+Kubernetes and containerd are required for the kubernetes cluster to start and should
+always be present, but others are can be added when required.
+
+NOTE: Only sha512 checksums are accepted by ignition.
```
m-bull, line 660: I think this has a proper .md notation - could swap to !!warning or something

.github/workflows/test.yaml
```diff
@@ -455,3 +455,90 @@ jobs:
...
+        if: ${{ vars.TARGET_CLOUD == 'arcus' }}
+
+      - name: Test Flatcar sysext deployment
+        uses: ./.github/actions/upgrade-and-test
+        with:
+          name: ci-${{ github.run_id }}-${{ github.job }}
+          os-cloud: ${{ vars.TARGET_CLOUD }}
+          chart-version: ${{ inputs.chart-version }}
+          kubernetes-version: ${{ vars.FLATCAR_KUBE_VERSION }}
+          image-id: ${{ vars.FLATCAR_IMAGE_ID }}
```
m-bull, line 520: This will be statically defined in repo vars - that's fine, but in which case we don't need to modification to ensure-image.sh.

Really I think I would prefer that this try and follow sort of the same image model as the Ubuntu side, but I can understand if you would rather fix that in a follow up

m-bull (changes requested):

PR #777, opened 2026-07-16
Fix broken dependency update CI

wtripp180901:
- Fixes missing .chart reference for amdGPUOperator in addon update CI
- Revert https://github.com/azimuth-cloud/capi-helm-charts/pull/754 and add tmpfs mount for /.cache to helm-unittest invocation

### DIFF
--- a/.github/workflows/update-addons.yml
+++ b/.github/workflows/update-addons.yml
@@ -183,11 +183,10 @@ jobs:
       - name: Update manifest snapshots
         run: |-
           helm dependency update charts/openstack-cluster && \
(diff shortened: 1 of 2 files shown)


irt-shpc (approved):
LGTM but can we trigger the nightly jobs that previously failed against this branch?

wtripp180901:
Looks like it works now
https://github.com/azimuth-cloud/capi-helm-charts/actions/runs/29500906075/job/87629292800
https://github.com/azimuth-cloud/capi-helm-charts/actions/runs/29500933432/job/87629397973

PR #803, opened 2026-07-21
Missing cloud config options from CAPI Helm Schema

JasleenKaurSethi:
Fix `cloudConfig` schema validation for values already supported by the `cloud-config.yaml` template.

The template supports some fields that the schema is missing causing Magnum CAPI Helm installs to fail when passing valid cloud config options.

Thus, 
- Added explicitly templated options to the schema
- additionalProperties true is set for fields which allow additional fields

### DIFF
--- a/charts/cluster-addons/values.schema.json
+++ b/charts/cluster-addons/values.schema.json
@@ -1789,11 +1789,11 @@
           "type": "object"
         },
         "cloudConfig": {
-          "additionalProperties": false,
+          "additionalProperties": true,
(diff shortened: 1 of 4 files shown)


wtripp180901 (changes requested):
Would be good to test the new fields by adding them to values_full.yaml, where that is possible.

Also these changes get overwritten when running `helm schema`, I think you can pass information for maps which don't have defaults set in values.yaml by defining them in schema annotations e.g
```
# @schema
# type: [object, "null"]
# properties:
#   host:
#     type: string
#   port:
#     type: integer
# @schema
myField:
```

wtripp180901 (approved):
LGTM

PR #805, opened 2026-07-22
feat: Add default 21 days on controlplane cert expiry

jihyun-huh:
## Summary

Set default value of `certificatesExpiryDays` to `21` in the CAPI Helm chart.

with this change, controlplane machine will be automatically rolled out `21` days before kubeadm certificates expire.

## Relationships with other PRs

This change follows the instruction of:
-  https://github.com/azimuth-cloud/ansible-collection-azimuth-ops/pull/1316
- https://github.com/azimuth-cloud/ansible-collection-azimuth-ops/pull/1317

## Backwards compatibility

Existing azimuth clusters that explicitly configure `certificateExpiryDays` would not be affected.
Users can override the default value through their helm values if a different rollout threshold is needed.

### DIFF
--- a/charts/openstack-cluster/tests/__snapshot__/snapshot_base_test.yaml.snap
+++ b/charts/openstack-cluster/tests/__snapshot__/snapshot_base_test.yaml.snap
@@ -926,6 +926,8 @@ templated manifests should match snapshot:
         minHealthyPeriod: 1h
         retryPeriod: 20m
       replicas: 3
+      rolloutBefore:
+        certificatesExpiryDays: 21
       rolloutStrategy:
         rollingUpdate:
           maxSurge: 1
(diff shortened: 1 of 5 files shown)


charts/openstack-cluster/values.schema.json
```diff
@@ -3267,7 +3267,8 @@
           "required": []
         },
         "certificatesExpiryDays": {
-          "description": "Roll out control plane Machines before their kubeadm-managed certificates expire. If unset, certificate-expiry based rollout is disabled.",
+          "default": 21,
```
m-bull: I don't think Helm respects this, I think it should be set in `values.yaml`.

If we had truly set a default then we would expect https://github.com/azimuth-cloud/capi-helm-charts/blob/c9aa00bee74466b6f043cfb83b98d9b89f1b5afc/charts/openstack-cluster/tests/__snapshot__/snapshot_full_test.yaml.snap#L2181 to also need updating. This is tested during the Lint CI job.

m-bull (changes requested):

charts/openstack-cluster/values.schema.json
jihyun-huh: @m-bull thanks! now i understand the structure.. i fixed it!

m-bull (approved):

PR #807, opened 2026-07-22
Add Helm values schema to all charts (pt2)

m-bull:
* charts/: Generate chart schemas for all helm charts

Schemas generated using https://github.com/dadav/helm-schema by running the following commands from the repo root:
```
helm plugin install https://github.com/dadav/helm-schema
helm schema
```
and then refining the generated schema using @schema comments in the relevant values.yaml files.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,77 @@
 # Contributing
 
+## Making a contribution
+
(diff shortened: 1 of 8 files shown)


wtripp180901:
Needs testing e2e in capi-operator before we merge

PR #809, opened 2026-07-27
feat: Add monitor kube-api cert expiry with blackbox exporter

jihyun-huh:
## Short Summary

This PR adds HA management Kubernetes API target(in-cluster) to the blackbox exporter, allowing a cluster to monitor the expiry of its own `kube-apiserver` certificate.

Background: This complements the Machine based monitoring added for workload clusters. In HA structure of Azimuth, the management cluster Machines exist on the seed cluster, while `kube-prometheus-stack` runs inside the management cluster. As a result, the management cluster cannot use its local `kube-state-metrics` instance to monitor its own controlplane Machines.

From the added blackbox rule, the new target probes `https://kubernetes.default.svc/livez` using the cluster CA from the `kube-root-ca.crt` ConfigMap. so the existing generic blackbox certificate alerts automatically apply to the resulting `target="kube-api"` metrics.

A kube-apiserver's specific critical alert is also added for certificates with 20 days, since `KubeadmControlPlane(KCP)` replacement should start at the 21 day expiry threshold, this indicates that automatic replacement may not have completed. 

The rule uses the minimum expiry observed over ten minutes to account for different controlplane backends being selected by the Kubernetes Service.

## Testing

* Ran `helm lint` for the `cluster-addons` chart
* Rendered `openstack-cluster` chart with the test values
* Updated and ran the Helm snapshot tests

### DIFF
--- a/charts/cluster-addons/values.yaml
+++ b/charts/cluster-addons/values.yaml
@@ -323,6 +323,9 @@ monitoring:
         #   targets:
         #     - name: example
         #       url: http://example.com/healthz
+    # Indicates if a blackbox target for the in-cluster Kubernetes API should be created
+    kubernetesApiTarget:
+      enabled: true
     # Indicates if a blackbox target for the cloud's auth endpoint should be created
     openstackAuthUrlTarget:
       enabled: true
--- a/charts/cluster-addons/templates/monitoring/blackbox-exporter.yaml
+++ b/charts/cluster-addons/templates/monitoring/blackbox-exporter.yaml
@@ -8,21 +8,54 @@ metadata:
     {{- include "cluster-addons.componentLabels" (list . "blackbox-exporter") | nindent 4 }}
     addons.stackhpc.com/watch: ""
 stringData:
-  {{- if .Values.monitoring.blackboxExporter.openstackAuthUrlTarget.enabled }}
+  {{- if or .Values.monitoring.blackboxExporter.kubernetesApiTarget.enabled .Values.monitoring.blackboxExporter.openstackAuthUrlTarget.enabled }}
   defaults: |
+    {{- if .Values.monitoring.blackboxExporter.kubernetesApiTarget.enabled }}
+    # Mount the cluster CA so that the in-cluster Kubernetes API endpoint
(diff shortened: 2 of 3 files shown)


charts/cluster-addons/values.yaml
```diff
@@ -317,12 +317,38 @@ monitoring:
       version: 11.9.1
     release:
       namespace: monitoring-system
-      values: {}
-        # Example of adding additional scrape targets
-        # serviceMonitor:
-        #   targets:
-        #     - name: example
-        #       url: http://example.com/healthz
+      values:
```
m-bull: I think this would work slightly more nicely in here: https://github.com/azimuth-cloud/capi-helm-charts/blob/d7ffc758b1369b49f6f7a68a1b6fd94ccc45b214/charts/cluster-addons/templates/monitoring/blackbox-exporter.yaml#L12 and put behind a config switch here in `values.yaml`, like the config for monitoring openstackAuthUrlTarget slightly below.

Otherwise I think the change is good.

m-bull (changes requested):

charts/cluster-addons/values.yaml
jihyun-huh: @m-bull  thanks for your review! 
I moved the values with several conditional statements, as to use both openstack-url-target and kube-api values.

charts/cluster-addons/values.yaml
```diff
@@ -324,6 +324,8 @@ monitoring:
         #     - name: example
         #       url: http://example.com/healthz
     # Indicates if a blackbox target for the cloud's auth endpoint should be created
+    kubernetesApiTarget:
```
m-bull, line 327: Please move the comment on L326 to L329 and add a comment describing kubernetesApiTarget

m-bull (changes requested):

charts/cluster-addons/values.yaml
jihyun-huh: oops, sorry
i fixed it!

m-bull (approved):
LGTM

PR #828, opened 2026-08-06
Dependency Updates 08/26

amanning9:


charts/openstack-cluster/tests/__snapshot__/snapshot_full_test.yaml.snap
```diff
@@ -1101,7 +1101,7 @@ templated manifests should match snapshot:
       chart:
         name: kube-prometheus-stack
         repo: https://prometheus-community.github.io/helm-charts
-        version: 85.0.2
+        version: 88.1.5
```
jihyun-huh: this bumps across three major chart versions (from 85 to 88), so my concern is, 
that Helm does not upgrade CRDs installed from the chart's `crds/` directly. but i don't think Azimuth enables `crds.upgradeJob.enabled`.

therefore, even if chart upgrade succeeds, the existing prometheus operator CRDs might not be upgraded. can we explicitly handle the CRD upgrade as part of this change? you can find the version boundaries here: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/UPGRADE.md

jihyun-huh (commented):
LGTM without the blocking concern

charts/openstack-cluster/tests/__snapshot__/snapshot_full_test.yaml.snap
amanning9: I've dropped this upgrade from this batch for now. Will install this upgrade separately, conversation continues here: https://github.com/azimuth-cloud/capi-helm-charts/pull/827

PR #831, opened 2026-08-07
Test Noble images in CI

m-bull:
This will never pass PR CI because of pull_request_target, but it would be useful to get this onto `main` for testing


wtripp180901 (approved):
LGTM

PR #834, opened 2026-08-11
fix: don't rely on OCCM for initializing nodes

Banh-Canh:


charts/openstack-cluster/templates/_helpers.tpl
```diff
@@ -330,6 +330,21 @@ preKubeadmCommands:
...
+Sets the kubelet cloud-provider arg on the given nodeRegistration configs.
+*/}}
+{{- define "openstack-cluster.cloudProvider" -}}
+{{- $provider := (index . 0).Values.kubeletCloudProvider }}
+{{- if $provider }}
+{{- range $config := rest . }}
+{{ $config }}:
+  nodeRegistration:
+    kubeletExtraArgs:
+      cloud-provider: {{ $provider }}
```
m-bull, line 343: I think this is missing a key part:


```suggestion
      cloud-provider: {{ $provider }}
      provider-id: openstack:///'{{ instance_id }}'
```

Like from https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/10dfeed6faa3e550a40712dc8b137c12bd4de255/templates/cluster-template.yaml#L24, but noting the change in apiVersion of the kubeadmconfigtemplate, which changes the formatting of `kubeletExtraArgs`

m-bull (changes requested):

charts/openstack-cluster/templates/_helpers.tpl
Banh-Canh: this is because I wanted to leave the default to rely on ccm for this more generic chart and then set it in 
https://github.com/azimuth-cloud/azimuth-capi-operator/pull/663/changes

is it preferred to set it as default ?

PR #838, opened 2026-08-17
Fix NVIDIA GPU Operator generating incompatible containerd config

wtripp180901:
By default, NVIDIA Container Toolkit tries to generate a config drop-in based on `containerd config dump`, which outputs config migrated to its containerd version (currently v4), meaning an incompatible config file is generated for the containerd version in the image (v2.3.2 currently) and crashes containerd. Sets to always generate config based on version defined in config.toml.
https://github.com/NVIDIA/aicr/issues/1688


irt-shpc (approved):
LGTM

